PR #8, opened 2017-10-30
issue #7 - Shutdown ExecutorService on JVM shutdown.

wajda:
Fixes delays on spark app shutdown


wajda:
Let's try to follow the internal dev cycle as was agreed before (review, fix the target branch etc). And merge this PR implicitly from the intranet.

PR #41, opened 2018-05-17
Feature/save as table operation lineage

ugoa:


vackosar:
Would u use rebase feature to squash your commits into single one? Would you reference an issue id in the comment if there is any related issue?

vackosar:
@ugoa perfect, I suggest to first respond to the rest of the comments and only at the end do the squashing as I am not sure how well github review will handle the squashing.

ugoa:
@fmalpica-sbd Thanks for point out, I'd fixed on my local branch. Will update it when it's all together.

@vackosar I wonder if github could merge with `--no-ff` option?

fmalpica-sbd:
@ugoa Anyway, I tried this code but there is some problem when it stores it on Mongodb because the web UI does not load it.

PR #55, opened 2018-11-14
Feature/merge streaming poc to dev

vackosar:


persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/LineageDAOv5.scala
```diff
@@ -0,0 +1,217 @@
...
+  })
+}
+
+object MutableLineageUpgraderV5 {
+
+  import LineageDAOv5.{ binaryTypeHintStrategy => hints }
+
+  def upgradeLineage(lineage: DBObject): Unit = {
+    for (op <- lineage.get(Component.Operation.name).asInstanceOf[ju.List[DBObject]].asScala) {
+      val opType = getOperationType(op)
```
wajda: I'd say if you create a `getOperationType()` than there should also be corresponding `setOperationType()` or alike. 
This piece of code does both getting and setting of the type, so the implementation should be consistent in both directions.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
```diff
@@ -46,6 +47,7 @@ abstract class BaselineLineageDAO extends VersionedLineageDAO with Logging {
   protected lazy val subComponents: Seq[SubComponent] = SubComponent.values
 
   protected lazy val dataLineageCollection: DBCollection = getMongoCollectionForComponent(Component.Root)
+  protected lazy val eventCollection: DBCollection = getMongoCollectionForComponent(Component.Event)
```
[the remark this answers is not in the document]
vackosar: I can move `eventCollection` into `LineageDaoV5` along with implementaion of methods like `getLineagesByPathAndInterval` and instead provide just dummy implementation so that MultiVersionDao can stay to be version-non-aware.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/LineageDAOv4.scala
```diff
@@ -51,6 +52,10 @@ class LineageDAOv4(override val connection: MongoConnection) extends BaselineLin
     super.save(lineage)
   }
 
+  override def saveProgress(event: ProgressDBObject)(implicit e: ExecutionContext): Future[Unit] = {
```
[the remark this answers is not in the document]
vackosar: will add dummy throwing exception in `BaseDao` and override that in `DaoV5` with above. 

Adding yet another interface is not a good idea. It may have been better to not to have MultiVersionDao at all and instead call from `MongoReader` n `MongoWriter` always directly `latestDao`, which would be responsible for calling the previous one. And so on. (daov5 would call daov4 then dao4 would call dao3).

But it is not worth changing now.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/LineageDAOv5.scala
vackosar: `getOperationType` extracts simple class name string of operation similarly to `MutableLineageUpgraderV4#getOperationType`.  Do you suggest to extract rest of the code which sets encoded field t into `setOperationType`?

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/LineageDAOv4.scala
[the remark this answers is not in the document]
wajda: Agree.
Regarding v5 to call v4 and so on I partially agree. API would be much clearer and separated, yes. But on the other side you'd have hard dependency between DAOs (v5 couldn't work without v4 etc) and each DAO would have to be responsible for upgrading the response from DAO vN-1 and combining the results. Though it would be possible to split the responsibility by implementing a chain-of-responsibility pattern, but I'm not sure if the final API would be clearer than the current one in the result.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/LineageDAOv5.scala
wajda: at least add a setter method. So if getType(x) == "Read", then to modify the type you call `setType(x, "Read"). 
In `DAOv4` there is `renameExpressionType` for instance.

wajda:
Also please add `spline-0.4.0.upgrade.js` to create all indices for `*_v5` collections.

web/pom.xml
```diff
@@ -237,10 +239,10 @@
                 <version>1.7</version>
                 <executions>
                     <execution>
-                        <id>ui-package</id>
-                        <phase>prepare-package</phase>
+                        <id>ui-download-dependencies</id>
+                        <phase>clean</phase>
```
wajda: Why on `clean` phase?

vackosar: A simple way not having to execute this everytime when building locally.

PR #58, opened 2018-11-23
Bugfix/spline 57 missing datatypes

wajda:


core/src/main/scala/za/co/absa/spline/core/transformations/LineageProjectionMerger.scala
```diff
@@ -60,7 +60,41 @@ object LineageProjectionMerger extends AsyncTransformation[DataLineage] {
...
+              case _ => accumulator
+            }
+            val updatedQueue = exp.children.toList ++ queue
+            traverseAndCollect(updatedAccumulator, updatedQueue)
+        }
+
+        traverseAndCollect(Set.empty, expressions.toList)
+      }
+
+      val dataTypeIds = {
```
vackosar: This variable name could be more descriptive e.g. usedDataTypes or retainedDataTypes.

wajda: renamed to `retainedDataTypes`

PR #72, opened 2018-12-10
Bugfix/#47

leborgneromain:
Fixed the bug on the sub-labels + refacto of the refreshHighlightedNodes function

### DIFF
--- a/web/ui/src/app/lineage/graph/graph-builder.ts
+++ b/web/ui/src/app/lineage/graph/graph-builder.ts
@@ -21,7 +21,7 @@ import * as _ from "lodash";
 import {OperationType, typeOfOperation} from "../types";
 import {VisModel} from "../../visjs/vis-model";
 import { ExpressionRenderService } from "../details/expression/expression-render.service";
-import { IJoin, IFilter, ISort, IRead } from "../../../generated-ts/operation-model";
+import { IJoin, IFilter, ISort, IRead, IWrite } from "../../../generated-ts/operation-model";
 
 export function lineageToGraph(lineage: IDataLineage,
                                expressionRenderService : ExpressionRenderService,
@@ -52,32 +52,40 @@ export function lineageToGraph(lineage: IDataLineage,
  * 
  * @returns the label from the operation
  */
-export function getLabel(operation : (IOperation), expressionRenderService : ExpressionRenderService){
-    let label = operation.mainProps.name + "\n"; 
+export function getLabel(operation : IOperation, expressionRenderService : ExpressionRenderService){
+    let label = operation.mainProps.name + "\n"
     switch(typeOfOperation(operation)){
         case "Aggregate":
-            let aggregations = (<any> operation).aggregations;
-            label += Object.keys(aggregations).join(",");break;
+            let aggregationExpressions = (<any> operation).aggregations
+            label += Object.keys(aggregationExpressions).join(",")
+            break
         case "Join":
-            let joins = (<IJoin> operation).condition;
-            label += expressionRenderService.getText(joins);break;
+            let joinExpression = (<IJoin> operation).condition
+            label += expressionRenderService.getText(joinExpression)
+            break
         case "Filter":
-            let filters = (<IFilter> operation).condition; 
-            label += expressionRenderService.getText(filters);break;
+            let filterExpression = (<IFilter> operation).condition
+            label += expressionRenderService.getText(filterExpression)
+            break
         case "Projection":
             // TODO : Define what to display for the second label
-            break;
+            break
         case "Sort":
-        let sorts = (<ISort> operation).orders; 
-        label += sorts[0].direction + " " + expressionRenderService.getText(sorts[0]);break;
+            let sortExpressions = (<ISort> operation).orders
+            label += sortExpressions[0].direction + " " + expressionRenderService.getText(sortExpressions[0])
+            break
         case "Read":
-            let read = (<IRead> operation).sources;
-            label += getFileName(read[0].path);break; 
+            let readExpression = (<IRead> operation).sources
+            label += getFileName(readExpression[0].path)
+            break
+        case "Write":
+            label += (<IWrite> operation).path
+            break
         default: 
-            label = operation.mainProps.name;
+            label = operation.mainProps.name
     }
 
-    return formalLabel(label);
+    return formatLabel(label)
 }
 
 /**
@@ -88,7 +96,7 @@ export function getLabel(operation : (IOperation), expressionRenderService : Exp
  * @returns the the filename of the path in parameters
  */
 function getFileName(path: string): string{
-    return path.replace(/^.*[\\\/]/, '');
+    return path.replace(/^.*[\\\/]/, '')
 }
 
 /**
@@ -98,12 +106,12 @@ function getFileName(path: string): string{
  * 
  * return the input label formated
  */
-function formalLabel(label: string, maxSize: number = 50): string{
-    label = label.substr(0, maxSize);
+function formatLabel(label: string, maxSize: number = 50): string{
+    label = label.substr(0, maxSize)
     if(label.length >= maxSize){
-        label += "...";
+        label += "..."
     }
-    return label;
+    return label
 }
 
 function createVisibleEdges(lineage: IDataLineage, hiddenOpIds: string[]): VisEdge[] {
--- a/web/ui/src/app/lineage/graph/graph-builder.ts
+++ b/web/ui/src/app/lineage/graph/graph-builder.ts
@@ -14,25 +14,25 @@
  * limitations under the License.
  */
 
-import {IAttribute, IDataLineage, IMetaDataset, IOperation} from "../../../generated-ts/lineage-model";
-import {RegularVisNode, VisEdge, VisNode} from "./graph.model";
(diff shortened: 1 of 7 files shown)


web/ui/src/app/lineage/graph/graph-builder.ts
```diff
@@ -34,13 +37,74 @@ export function lineageToGraph(lineage: IDataLineage,
...
+/**
+ * Get label from an operation
+ * 
+ * @param operation the operation where the label should be retrieved
+ * @param expressionRenderService service to render expressions
+ * 
+ * @returns the label from the operation
+ */
+export function getLabel(operation : (IOperation), expressionRenderService : ExpressionRenderService){
+    let label = operation.mainProps.name + "\n";
```
wajda: Semicolons. Unless you have a strong opinion why we _should_ use semicolons I'd say we rather don't.
While in JS world, there is a never ending battle about if semicolons are good or evil, in TS however they make even less sense. Even in JS the cases that make ASI to behave unexpectedly are rare and strange enough to not belong to a good coding style anyway. But in TS the compiler will catch a good chuck of those leaving even less room for ASI related mistakes. There is a good article about it - https://medium.com/@eugenkiss/dont-use-semicolons-in-typescript-474ccfe4bdb3

web/ui/src/app/lineage/graph/graph-builder.ts
```diff
@@ -34,13 +37,74 @@ export function lineageToGraph(lineage: IDataLineage,
...
+ * @param expressionRenderService service to render expressions
+ * 
+ * @returns the label from the operation
+ */
+export function getLabel(operation : (IOperation), expressionRenderService : ExpressionRenderService){
+    let label = operation.mainProps.name + "\n"; 
+    switch(typeOfOperation(operation)){
+        case "Aggregate":
+            let aggregations = (<any> operation).aggregations;
+            label += Object.keys(aggregations).join(",");break;
```
wajda: `break` should be on a new line.
Having it in the end of the previous line makes it very easy to overlook when reading the code. If you want to keep the `switch `-block  more compact consider moving it in a separate function and use `case ... : return ...` pattern, concatenating strings outside.

web/ui/src/app/lineage/graph/graph-builder.ts
```diff
@@ -34,13 +37,74 @@ export function lineageToGraph(lineage: IDataLineage,
...
+        case "Join":
+            let joins = (<IJoin> operation).condition;
```
wajda: Try to find more meaningful names for variables. The names like `joins` or `filters` don't make it obvious that they contain a value of a singular expression type.

web/ui/src/app/lineage/graph/graph-builder.ts
```diff
@@ -34,13 +37,74 @@ export function lineageToGraph(lineage: IDataLineage,
...
+            label += expressionRenderService.getText(joins);break;
+        case "Filter":
+            let filters = (<IFilter> operation).condition; 
+            label += expressionRenderService.getText(filters);break;
+        case "Projection":
+            // TODO : Define what to display for the second label
+            break;
+        case "Sort":
+        let sorts = (<ISort> operation).orders;
```
wajda: indentation

web/ui/src/app/lineage/graph/graph-builder.ts
vackosar: I prefer semicollons in TS, but I  dont mind either way.

leborgneromain: As you guys prefer, I have no preference.

web/ui/src/app/lineage/graph/graph-builder.ts
leborgneromain: ok ok. I am used to an autoformat on save that was obviously not here.

web/ui/src/app/lineage/graph/graph-builder.ts
leborgneromain: Renammed

web/ui/src/app/lineage/graph/graph-builder.ts
```diff
@@ -34,13 +37,74 @@ export function lineageToGraph(lineage: IDataLineage,
...
+            label += expressionRenderService.getText(filters);break;
+        case "Projection":
+            // TODO : Define what to display for the second label
+            break;
+        case "Sort":
+        let sorts = (<ISort> operation).orders; 
+        label += sorts[0].direction + " " + expressionRenderService.getText(sorts[0]);break;
+        case "Read":
+            let read = (<IRead> operation).sources;
+            label += getFileName(read[0].path);break;
```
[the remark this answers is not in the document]
leborgneromain: Added the path for the write operation.

web/ui/src/app/lineage/graph/graph-builder.ts
```diff
@@ -34,13 +37,74 @@ export function lineageToGraph(lineage: IDataLineage,
...
 }
+/**
+ * Get label from an operation
+ * 
+ * @param operation the operation where the label should be retrieved
+ * @param expressionRenderService service to render expressions
+ * 
+ * @returns the label from the operation
+ */
+export function getLabel(operation : (IOperation), expressionRenderService : ExpressionRenderService){
```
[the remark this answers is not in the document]
leborgneromain: yes they are. Removed

web/ui/src/app/lineage/graph/graph-builder.ts
```diff
@@ -34,13 +37,74 @@ export function lineageToGraph(lineage: IDataLineage,
...
         hiddenOpIds: string[] = _.map(hiddenOperations, "mainProps.id"),
-        visibleNodes = visibleOperations.map(op => new RegularVisNode(op)),
+        visibleNodes = visibleOperations.map(op => new RegularVisNode(op, getLabel(op, expressionRenderService))),
         visibleEdges = createVisibleEdges(lineage, hiddenOpIds)
 
     return new VisModel(
         new vis.DataSet<VisNode>(visibleNodes),
         new vis.DataSet<VisEdge>(visibleEdges))
 }
+/**
```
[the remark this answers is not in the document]
leborgneromain: Well, I am trying to comment every of my function even tho I do agree that these one are pretty straight forward functions.

leborgneromain:
Corrected code after your code review.

web/ui/src/app/lineage/graph/graph-builder.ts
```diff
@@ -34,13 +37,82 @@ export function lineageToGraph(lineage: IDataLineage,
...
+}
+
+/**
+ * Takes the first maxSize characters and add ellipsis to it
+ * @param label the label to format
+ * @param maxSize the size of the final string
+ * 
+ * return the input label formated
+ */
+function formatLabel(label: string, maxSize: number = 50): string{
```
wajda: This function is redundant. There is `truncate` method in LoDash that does exactly that. And it's also a little bit more accurate at placing "..." when the original string length is equal to the max length.
Example: `_.truncate(label, {length: maxSize})`

web/ui/src/app/lineage/graph/graph.model.ts
```diff
@@ -47,16 +47,24 @@ export abstract class VisNode implements vis.Node {
                 public icon: any,
                 public type: VisNodeType) {
     }
+
```
wajda: Another redundant method. Please take a look at LoDash `clone()` function. There are four variants of it.

web/ui/src/app/lineage/graph/graph.model.ts
```diff
@@ -70,11 +78,12 @@ export class RegularVisNode extends VisNode {
 }
 
 export class HighlightedVisNode extends VisNode {
-    constructor(public operation: IOperation) {
+    constructor(
+        public operation: IOperation, public label : string) {
```
wajda: Minor thing: Line breaks. For a multi-line method/constructor signature, it's a good practice to put every argument on its own line. So you either place everything on one line or one arg per line.

leborgneromain: corrected

web/ui/src/app/lineage/graph/graph.model.ts
leborgneromain: corrected

web/ui/src/app/lineage/graph/graph-builder.ts
leborgneromain: corrected

wajda:
Probably the last thing - since you added `package-lock.json` into the repo it should also be removed from the `.gitignore` file.

leborgneromain:
> Probably the last thing - since you added `package-lock.json` into the repo it should also be removed from the `.gitignore` file.

No no that should not be commited. The package-lock.json has nothing to do at the root of the project. this was a wrong npm command launched and stopped manually on this directory. Sorry.

wajda:
LGTM.

> > Probably the last thing - since you added `package-lock.json` into the repo it should also be removed from the `.gitignore` file.
> 
> No no that should not be commited. The package-lock.json has nothing to do at the root of the project. this was a wrong npm command launched and stopped manually on this directory. Sorry.

Actually, and quite surprisingly, according to the NPM doc it is supposed to be committed.
But that's not important for this PR. 

> This file is intended to be committed into source repositories, and serves various purposes:
> https://docs.npmjs.com/files/package-lock.json

PR #73, opened 2018-12-11
Support Spark 2.4 Streaming

vackosar:


spark-adapter-2.2/src/main/scala/za/co/absa/spline/sparkadapterapi/KafkaSinkAdapterImpl.scala
```diff
@@ -24,13 +25,15 @@ import scala.collection.JavaConverters._
...
-        val params = ReflectionUtils.getFieldValue[ju.Map[String, String]](x, "executorKafkaParams").asScala
-        val topics = ReflectionUtils.getFieldValue[Option[String]](x, "topic").get.split(",").toList
-        val bootstrapServers = params("bootstrap.servers").split("[\t ]*,[\t ]*")
-        Some(KafkaSinkInfo(topics, bootstrapServers))
-      case _ => None
+    val sink = streamExecution.sink
+    if (sink.getClass.getCanonicalName == "org.apache.spark.sql.kafka010.KafkaSink") {
+      val params = getFieldValue[ju.Map[String, String]](sink, "executorKafkaParams").asScala
+      val topics = getFieldValue[Option[String]](sink, "topic")
+        .map(_.split(",").toSeq).getOrElse(Seq())
```
wajda: for empty sequences, instead of `Seq()` it's better to use `Seq.empty` or `Nil`. It's just easier for eyes to grasp it.

PR #74, opened 2018-12-17
spline #71 - Create shadow Spline module

wajda:
Added a simple version of the Migrator Tool that reads from Mongo and writes to Arango.
Arango persistor is to be implemented in the next PR


vackosar:
Actor model is quite interesting!

PR #79, opened 2019-01-02
Bugfix/spline 52 ignore mode

wajda:


core/src/main/scala/za/co/absa/spline/core/listener/SplineQueryExecutionListener.scala
```diff
@@ -39,10 +40,16 @@ class SplineQueryExecutionListener(harvesterFactory: DataLineageBuilderFactory,
     if (funcName == "save") {
       log debug s"Start tracking lineage for action '$funcName'"
 
-      val builder = harvesterFactory.createBuilder(qe.sparkSession.sparkContext)
+      val rawLineage =
+        harvesterFactory.
+          createBuilder(qe.analyzed, Some(qe.executedPlan), qe.sparkSession.sparkContext).
```
vackosar: Upon merge to develop will method createBuilder needed? Seems like everything createBuilder could be just renamed to buildLineage. But right now you probably intend to reduce conflicts.

core/src/main/scala/za/co/absa/spline/core/listener/SplineQueryExecutionListener.scala
```diff
@@ -39,10 +40,16 @@ class SplineQueryExecutionListener(harvesterFactory: DataLineageBuilderFactory,
...
+          buildLineage()
 
-      val rawLineage = builder.buildLineage(qe.analyzed)
-      lineageProcessor.process(rawLineage)
+      if (wasResultIgnored(rawLineage)) {
+        log debug s"The write result was ignored. Skipping lineage."
```
vackosar: So here we will probably place creation of empty Progress object in future.

core/src/main/scala/za/co/absa/spline/core/listener/SplineQueryExecutionListener.scala
```diff
@@ -51,6 +58,14 @@ class SplineQueryExecutionListener(harvesterFactory: DataLineageBuilderFactory,
     }
   }
 
+  def wasResultIgnored(lineage: DataLineage): Boolean =
+    lineage.rootOperation match {
+      case op.Write(_, _, _, _, writeMetrics, readMetrics) =>
+        writeMetrics get "numFiles" exists 0.==
+      case _ =>
+        sys.error(s"Unexpected root operation: ${lineage.rootOperation.getClass}")
```
vackosar: Why not rather use "warning" or throw an exception? Error usually means that the program cannot continue.

wajda: Because the program cannot continue if this happens :) This code expects the root operation to always be `Write`. Basically I could safely remove `case _`  block and it won't change anything. I have put it there to emphasize and describe this assumption. Another way to do it would be `assume(rootOperation.isInstanceOf[Write])`, but `case _ => sys.error` reads better to me.

wajda: In my other projects I usually had a common helper object 'Assert` with methods like `shouldNeverReachHere()`, `shouldNeverBeThrown(ex)` etc that would simply throw a runtime exception with the according description.

core/src/main/scala/za/co/absa/spline/core/listener/SplineQueryExecutionListener.scala
vackosar: Interesting. nitpick: logicalPlan could have stayed as buildLineage function parameter as it is used only there.

core/src/main/scala/za/co/absa/spline/core/listener/SplineQueryExecutionListener.scala
vackosar: Oh, sys.error actually throws the RuntimeException.

PR #81, opened 2019-01-07
Feature/spline2 ui

leborgneromain:


leborgneromain:
The codacy checks will be commit on the next pull request.

Issue #75, opened 2018-12-19
Return Atlas support to Spline 0.3


wajda:
A lot of people asked us to return Atlas support into a stable Spline 0.3.x version. So here we go.

vackosar:
@rjbarrington @xbbnrde  @vjbhakuni2 @sanjurm16 @ssenathi

Can you **help us testing Atlas support** that was just **reintroduced** onto branch **release/0.3** by the end of next week? Issues reported util the end of next week will have higher chance of being fixed as we will still have the original developer available.

The supported Atlas version is 1.0. To test you need to checkout the branch and build Spline snapshot version locally. If you don't have Atlas 1.0 setup, there is a step by step guide how to do it on announcement blog post https://absaoss.github.io/spline/2019/01/24/Atlas-Support-Is-Back.html

Let us know here please!

vackosar:
(Atlas support was released in 0.3.6)

PR #92, opened 2019-01-23
Support for Apache Atlas 1.0

mn-mikke:
The PR solves issue #75 by adding persistence layer to Apache Atlas 1.0.


persistence/atlas/src/main/atlas/spline-meta-model.json
```diff
@@ -0,0 +1,604 @@
+{
```
vackosar: Is it recognizable from the Atlas UI that the lineage was published by Spline?

vackosar: How compatible within Atlas is this with say NifiLineageReporter? I think this is its meta model: https://github.com/vakshorton/NifiAtlasLineageReporter/blob/master/AtlasTypes/nifiTypes.json

mn-mikke: > Is it recognizable from the Atlas UI that the lineage was published by Spline?

Explicitly no, but you can see all types defined in this meta-model in Atlas UI and those refer to Spark. IMHO, an explicit information about Spline doesn't give a user any additional value. WDYT?
>  How compatible within Atlas ...

see specification of ```FileEndpoint``, it's published as a built-in Atlas Type (hdfs_path), so if any other tool produce an entity with the same URI, the lineage graphs will join.

PR #96, opened 2019-01-24
Feature/spline 86 webjar

wajda:


spline-shadow/client-web/src/main/scala/za/co/absa/spline/client/web/WebController.scala
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.stereotype.Controller
+import org.springframework.web.bind.annotation.{RequestMapping, ResponseBody}
+import org.webjars.WebJarAssetLocator
+import za.co.absa.spline.common.SplineBuildInfo
+
+@Controller
+class WebController @Autowired()(webJarAssetLocator: WebJarAssetLocator) {
+
+  @RequestMapping(path = Array("/"))
+  def root(httpRequest: HttpServletRequest): String = s"redirect:/absaoss-spline-client/spline/"
```
vackosar: This is temporary mapping? I am not sure what it would do in prod.

spline-shadow/rest-gateway/src/main/scala/za/co/absa/spline/gateway/rest/controller/AboutController.scala
```diff
@@ -16,23 +16,21 @@
...
-import scala.io.Source.fromInputStream
+import za.co.absa.spline.common.SplineBuildInfo
 
 @RestController
 @RequestMapping(Array("/about"))
 class AboutController {
 
-  @GetMapping(path = Array("/build"), produces = Array("text/x-java-properties"))
+  @GetMapping(path = Array("/build"), produces = Array(APPLICATION_JSON_VALUE))
   @ApiOperation("Get application version and build information")
```
vackosar: You could name the endpoint appVersionAndBuildInfo and then the comment would be redundant

wajda: This is not a comment, it's part of Swagger documentation. It's supposed to be verbose.
API itself is already concise and self explanatory - /about/build

Issue #109, opened 2019-02-05
Remove unnecessary dependencies to Kafka client from Spline core and API module


wajda:

wajda:
@yruslan

PR #110, opened 2019-02-05
Removing streaming and all kafka dependencies from Spline 0.3.x

mn-mikke:
The PR solves issue #109 and removes all dependencies to spark-sql-kafka-0.10 and all code related to streaming.


vackosar:
We need to be careful not to actually delete the files on develop branch as well with the merge.

wajda:
I'll merge it manually

PR #128, opened 2019-02-17
Refactoring/apply spark fixture in tests

wajda:
Fixes some problems related to Spark context and session re-usage during tests execution.


harvester/src/test/scala/za/co/absa/spline/harvester/SampleJobSpec.scala
```diff
@@ -32,6 +32,9 @@ import scala.collection.mutable.ArrayBuffer
 import scala.concurrent._
 import scala.concurrent.duration.Duration
```
vackosar: it tests there is no exception

harvester/src/test/scala/za/co/absa/spline/harvester/SparkLineageInitializerSpec.scala
```diff
@@ -94,108 +87,117 @@ class SparkLineageInitializerSpec extends FunSpec with BeforeAndAfterEach with M
...
+          assertSplineIsDisabled()
+        })
       }
 
       it("should abort application, when is in REQUIRED mode") {
         intercept[Exception] {
-          configuration.setProperty(MODE, REQUIRED.toString)
-          val session = createSession
-          session.enableLineageTracking(createFailingConfigurer(session))
+          withNewSession(session => {
```
vackosar: why not use something like @before that is used in JUnit

wajda: those are alternative techniques. In fact in scalatests you can choose between three different techniques for implementing fixtures. The one that uses 'before' and 'after' methods is useful when you want to have more fine-grained control on the fixture init and release steps. And it also aborts the entire suite in case of an error during a fixture initialization. So before-and-after is a more heavy-weight technique.
The one that you pointer (`withNewsession`-method) is called a _loan-pattern_. It allows you to decide on the fixture details on a per test method basis. It is more suitable for tests when each methods might need different state of a set of fixtures.
Basically `SparkFixture` implements both patterns and can be used in both ways.

PR #130, opened 2019-02-18
Persister refactoring, simplify SerDe, unify Arango driver usage

wajda:
- Remove scarango dependency.
- Add velocypack-module-scala.
- Merge "model" module into "persistence" module.


spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/DataLineageTransactionParams.scala
```diff
@@ -18,135 +18,114 @@ package za.co.absa.spline.persistence
...
+
+  def fields: Array[String] = getClass
+    .getDeclaredFields
+    .map(_.getName)
+    .filter(_ != "$outer")
+
+  def saveCollectionsJs: String = fields
+    .map(field =>
+      s"""
+         |  params.$field.forEach(o => db.$field.insert(o));
```
vackosar: In this case you should be able to do ``` db.$field.insert(params.$field) ```

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
```diff
@@ -0,0 +1,89 @@
...
+trait Operation extends PolymorphicDocumentOption {
+  def name: String
+
+  def expression: String
+
+  def outputSchema: Schema
+}
+
+case class Read(name: String, expression: String, format: String, outputSchema: Schema, _key: Option[String] = None, _id: Option[String] = None, _rev: Option[String] = None, _type: String = "read") extends Operation {
+  def this() = this("", "", "", null)
```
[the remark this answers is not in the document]
wajda: Yes. Unfortunately __velocypack-module-scala__ while teaching __java-velocypack__ how to cope with scala types, it doesn't alter the underlying reflection mechanism, which (being originally written for java) relies on the Java Bean convention.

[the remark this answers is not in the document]
wajda: A good thing however is that it doesn't require `@BeanProperty` annotations on every field, which I would normally expect. Apparently it's not fully Java Bean compliant and can reflect fields directly.

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/DataLineageTransactionParams.scala
wajda: I tried :) but it didn't work. In JS in order to use method as lambda, provided that the method needs to have access to `this`, you'd need to bind the method to the object explicitly.

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/DataLineageTransactionParams.scala
```diff
@@ -18,135 +18,114 @@ package za.co.absa.spline.persistence
...
-      createReadsFrom(dataLineage, uriToKey).asJava,
-      createApp(dataLineage).asJava,
-      createImplements(dataLineage).asJava,
-      createExecution(dataLineage).asJava,
-      createExecutes(dataLineage).asJava,
-      createProgressForBatchJob(dataLineage).asJava,
-      createProgressOf(dataLineage).asJava
+  def create(dataLineage: DataLineage, uriToNewKey: Map[String, String], uriToKey: Map[String, String]): DataLineageTransactionParams = {
+
+    implicit def ser(it: scala.collection.Iterable[AnyRef]): Iterable[VPackSlice] = it.map(Persister.vpack.serialize).asJava
```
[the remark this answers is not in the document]
wajda: I don't now. I didn't went all the way down with the refactoring. So there might be left overs.

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/DataLineageTransactionParams.scala
vackosar: What I mean is that insert method accepts also arrays and not only single values.

vackosar: Actually I am not sure if ```insert``` can do it but ```save``` should be able to.

wajda: Oh, yes, sorry, I misread your comment (thought it were `...forEach(db.field.insert)`)
Regarding `insert(array)` I tried it as well. The problem with it was that this method doesn't throw exceptions. Instead it returns an array of results per item. But since we need to fail fast and abort the transaction on an error in any item I thought `forEach(o => insert(o))` does better.

wajda: According to the doc `insert` is just an alias to `save`, so no matter which one to use.

PR #138, opened 2019-02-28
Feature/spline 119 rest get exec plan

wajda:


spline-shadow/rest-core/pom.xml
```diff
@@ -72,7 +72,6 @@
         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-webmvc</artifactId>
-            <version>5.1.3.RELEASE</version>
```
leborgneromain: The version 5.1.5.RELEASE is also available.

spline-shadow/rest-core/src/main/scala/za/co/absa/spline/gateway/rest/model/ExecutionInfo.scala
```diff
@@ -16,18 +16,19 @@
...
-  nativeId: Option[String],
-  startTime: Option[Date],
-  finishTime: Option[Date]
+  _type: String,
+  startTime: Option[Long],
+  endTime: Option[Long],
+  extra: Map[String, Any]
 ) {
-  require(Seq(startTime, finishTime).exists(_.isDefined), "at least one of 'start' or 'finish' time should be defined")
+  def this() = this(null, null, null, null, null)
```
leborgneromain: I am not familiar with scala but is it necessary to define such a constructor ?

vackosar: It is probably needed for the json serialization

spline-shadow/rest-core/src/main/scala/za/co/absa/spline/gateway/rest/model/ExecutedLogicalPlan.scala
```diff
@@ -22,24 +22,31 @@ import za.co.absa.spline.gateway.rest.model.ExecutedLogicalPlan._
...
+  plan: LogicalPlan
+) {
+  def this() = this(null, null)
+}
 
 object ExecutedLogicalPlan {
   type OperationID = UUID
 
-  case class LogicalPlan(nodes: Seq[Operation], edges: Seq[Transition]) extends Graph {
+  case class LogicalPlan(nodes: Array[Operation], edges: Array[Transition]) extends Graph {
```
vackosar: Is this generic enough name? I mean, I don't think logical plan has good equivalent in StreamSets.

spline-shadow/rest-core/src/main/scala/za/co/absa/spline/gateway/rest/model/ExecutionInfo.scala
wajda: Yes, unfortunately it's one of those scala-java edge stones. Arango Java driver require model classes to have a default constructor (the one with no args), that is one of fundamental requirement in Java bean spec. But Scala case classes doesn't follow Java bean spec and do not generate no-arg constructors by default.
I have a plan to write a macro activated by a static annotation that would generate default constructor for an annotated case class. But that's not an easy task, so I saved it for later.

spline-shadow/rest-core/src/main/scala/za/co/absa/spline/gateway/rest/repo/ExecutionPlanRepository.scala
```diff
@@ -14,10 +13,15 @@
...
 
-package za.co.absa.spline.gateway.rest.model
+package za.co.absa.spline.gateway.rest.repo
+
+import za.co.absa.spline.gateway.rest.model.ExecutedLogicalPlan
+import za.co.absa.spline.gateway.rest.model.ExecutionInfo.Id
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait ExecutionPlanRepository {
```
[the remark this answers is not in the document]
wajda: First, it has at least two implementations - one is static (Arango one) and the other is a dynamic mock used in the rest-api-doc-gen module.
Second, actually it is a good thing to have traits between layers (in this case it's a controller-dao boundary) even if there is only one impl. It is called an Interface-based design and it has many advantages. One of the most practical and straightforward one is testability.

spline-shadow/rest-gateway/src/main/scala/za/co/absa/spline/gateway/rest/ArangoRepoConfig.scala
```diff
@@ -0,0 +1,79 @@
...
+  setThrowExceptionOnMissing(true)
+
+  override val rootPrefix: String = "spline"
+
+  object Database extends Conf("database") {
+
+    // todo: allow symbols '@' and ':' in passwords
+    private val ArangoConnectionUrlRegex =
+      ("arangodb://"
+        + "(?:"
```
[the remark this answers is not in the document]
wajda: thx, will try.

spline-shadow/rest-gateway/src/main/scala/za/co/absa/spline/gateway/rest/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -0,0 +1,61 @@
...
+
+@Repository
+class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends ExecutionPlanRepository {
+
+  override def findById(execId: Id)
+                       (implicit ec: ExecutionContext): Future[ExecutedLogicalPlan] = {
+    db.queryOne[ExecutedLogicalPlan](
+      """
+      FOR exec IN execution
+          FILTER exec._key == CONCAT("ln_", @execId)
```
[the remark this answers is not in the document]
wajda: oops. how could it make it there? This was removed in the commit 169a1c8b230696dc07f119f00b143a2ee297cbf0. GitHub glitch?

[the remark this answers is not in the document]
wajda: Oh, it says "Outdated". That's probably because I made a several pushes.

spline-shadow/rest-core/src/main/scala/za/co/absa/spline/gateway/rest/model/ExecutedLogicalPlan.scala
wajda: Maybe you're right. What about just ExecutionPlan? I think you can't really get more generic than that. The only doubt that I have is that if we happen to also store a physical plan (and we forget about YAGNI)  then the work "logical" here would help to avoid confusions.

spline-shadow/rest-core/src/main/scala/za/co/absa/spline/gateway/rest/repo/ExecutionPlanRepository.scala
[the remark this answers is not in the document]
vackosar: For mocking u dont need the interface. Yes, I know this approach, but I think it is just snakeoil. Java class gives u an interface - public methods.

spline-shadow/rest-core/src/main/scala/za/co/absa/spline/gateway/rest/model/ExecutedLogicalPlan.scala
vackosar: exec plan sounds good or executionLineage.

PR #139, opened 2019-03-01
saveAsTable support

lukasnalezenec:


vackosar:
I restarted your build. It probably failed when I was experimenting with making Mongo docker set to use replication.

vackosar:
Good job!
Best is to have your commit messages referencing your issue. Also rebasing squashing commits is nice.

PR #141, opened 2019-03-01
Feature/ui2

leborgneromain:


spline-shadow/client-ui/src/app/components/lineage/lineage-details/schema-table/schema-table.component.ts
```diff
@@ -13,78 +13,94 @@
...
-  rowClick(item: any) {
-    this.propertyService.changeCurrentProperty(item)
-    this.selectedItemName = item.name
+  /**
+   * Gets selected row from name
+   * @param name the name of the property
+   * @returns a tuple containing the row itself and it's index in case the table is pageable
+   */
+  getSelectedRowFromName(name) {
+    let selectedRow = null
```
wajda: Looks like the method is just `const i = _.findLastIndex(rows, {name: name}); return [i, rows[i]]`.
If in fact there is no duplicated row names then using `_.findIndex` will be faster as it terminates earlier.

spline-shadow/client-ui/src/app/components/lineage/lineage-graph/lineage-graph.component.ts
```diff
@@ -48,11 +53,28 @@ export class LineageGraphComponent implements OnInit {
...
         that.cytograph.cy.add(response)
 
-        that.cytograph.cy.nodes().on('click', function (event) {
-          let clickedNode = event.target
-          that.lineageGraphService.getDetailsInfo(clickedNode.id())
+        that.cytograph.cy.on('click', function (event) {
+          const clikedTarget = event.target
+          let nodeId = null
+          if (clikedTarget != that.cytograph.cy && clikedTarget.isNode()) {
+            nodeId = clikedTarget.id()
```
wajda: Please prefer more functional without reassignment. E.g. `const nodeId = `clikedTarget != that.cytograph.cy && clikedTarget.isNode() ? clikedTarget.id() : null` if you like.

PR #142, opened 2019-03-06
#119 UI Integration of the getExecutedLogicalPlan service

leborgneromain:


spline-shadow/rest-gateway/src/main/scala/za/co/absa/spline/gateway/rest/AppConfig.scala
```diff
@@ -16,9 +16,45 @@
...
+import za.co.absa.spline.gateway.rest.CorsConfig.CorsFilterConf
 
 @Configuration
 @Import(Array(classOf[ArangoRepoConfig]))
-class AppConfig
+class AppConfig{
+
+
+
+  val source: UrlBasedCorsConfigurationSource  = new UrlBasedCorsConfigurationSource()
```
wajda: Lines 35-44 should move to  the `springFilterPoxy` method because the `AppConfig` is supposed to contain different loosely related bean definitions. So there should only be `@Bean def .....` on the class members level.

spline-shadow/rest-gateway/src/main/scala/za/co/absa/spline/gateway/rest/AppConfig.scala
```diff
@@ -16,9 +16,45 @@
...
+    case _ =>  config.addAllowedHeader("*")
+  }
+  config.addAllowedMethod("GET")
+  source.registerCorsConfiguration("/**", config)
+
+  @Bean def springFilterProxy = new CorsFilter(source)
+}
+
+
+object CorsConfig extends CompositeConfiguration(asList(
```
wajda: `CorsConfig` should be places into a separate file. Usually (excluding some specific cases) in Scala only companion objects are placed in the same file with the corresponding class.

vackosar:
I just noticed rest-core module contains code in package "spline.gateway.rest". Usually module's code is in a package corresponding to its name eg "spline.restcore". It is like this because of need of cooperation with rest-gateway module?

wajda:
> I just noticed rest-core module contains code in package "spline.gateway.rest". Usually module's code is in a package corresponding to its name eg "spline.restcore". It is like this because of need of cooperation with rest-gateway module?

I don't think it's related. Most likely the package name just wasn't updated during the module split. Ignore it for now, we'll rename packages when we have a stable module names. I expect a few more renames during next couple of weeks. The problem is that package rename is a heavy operation in java and git as it requires physically moving the files between directories and may seriously confuse the history and the mess up merges. So it's better not to do it often.

Issue #21, opened 2018-02-15
"saveAsTable" and Spark Catalog support


wajda:

ugoa:
Hi @wajda,

First of all big thanks for your work on the Spline project, it's has been a great help to us. I was wondering if and how I could contribute to this project to help it growing mature, especially for the HIVE interaction part. Appreciate it if you can provide any advices. Cheers!

wajda:
Hi @ugoa,
Thank you for your message. Absolutely, you can contribute and every contribution will be appreciated. We don't have a strict documented procedure at the moment, but in a nutshell the process is the following:
1. You fork the Spline repo and do whatever changes you want.
2. When you feel ready to contribute the changes back simply create a PR to the main repo, master branch, and explain briefly what changes you are proposing, what it's for and how it's implemented and tested.
3. We'll review your PR, test it and when we are satisfied with the changes we'll finally merge it into our internal upstream repo. The merged changes will be published back on GitHub with the next Spline release.

Unfortunately due to our current company restrictions we cannot develop directly on GihHub, so we only publish released versions from time to time. This should change soon however.

Let's keep in touch.
Regards, Alex.

ugoa:
Hi Alex,

Glad to hear you back. I'd just started to dig into the code, and I was wondering if there were any sharable documents that can help me for better understanding and developing the project? If so that would be terrific. Thanks!

Dawei, Regards.

wajda:
Unfortunately there is none yet. The project is still considered PoC, many changes are still coming. Please feel free to email me or Vaclav for any questions or topics to discuss about the project.

ugoa:
Oleksandr.Vayda@barclayscapital.com. Is it correct? Surely I will try to reach you when I got questions. Cheers.

wajda:
Yes, that's correct

vackosar:
@ugoa , Absa is using this code internally and we are responsible for stability and maintainability. So I suggest to discuss first what you would like to contribute as we will have some requirements for what we will merge into this repo.

@ugoa, do you have already something in mind that you would like to do?

ugoa:
Hi @vackosar, I am interested in Spark reading/saving tables with Hive, I was wondering if your team already have a plan for this, or in what way I could provide some help.

And for now I am still reading the source code and playing with it.

sanjurm16:
[...]

PR #143, opened 2019-03-07
spline #21 saveAsTable support

lukasnalezenec:


vackosar:
This is replacement for PR https://github.com/AbsaOSS/spline/pull/139 ? That one will be closed?

core/src/main/scala/za/co/absa/spline/core/harvester/DataLineageBuilder.scala
```diff
@@ -25,29 +25,42 @@ import org.apache.spark.sql.execution.datasources.LogicalRelation
...
 
+  private val writeCommandParser = writeCommandParserFactory.getWriteParser()
+  private val tableCommandParser = writeCommandParserFactory.getSaveAsTableParser()
+  private val jdbcCommandParser = writeCommandParserFactory.getJDBCParser()
+
   def buildLineage(): DataLineage = {
+    val builders = getOperations(logicalPlan)
+    val someRootBuilder = builders.lastOption
+
+    val writeIgnored = if (someRootBuilder.isDefined && someRootBuilder.get.isInstanceOf[RootNode]) {
```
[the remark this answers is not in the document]
wajda: Definitely should be done with pattern-matching. Every time when patterns like the following appear it's a case for pattern-matching. This piece of code has both at a time:
```scala
if (monad.isDefined) monad.get
if (x.isInstanceOf[T]) x.asInstanceOf[T]
if (list.notEmpty) list.get(0)
etc etc
```

core/src/main/scala/za/co/absa/spline/core/harvester/DataLineageBuilder.scala
```diff
@@ -62,11 +75,25 @@ class DataLineageBuilder(logicalPlan: LogicalPlan, executedPlanOpt: Option[Spark
...
+            val writes = writeCommandParser.
                 asWriteCommandIfPossible(curOpNode).
                 map(wc => Seq(wc.query)).
-                getOrElse(curOpNode.children)
+                getOrElse(Seq())
+
+            val tables = tableCommandParser.
+              asWriteCommandIfPossible(curOpNode).
+              map(wc => Seq(wc.query)).
+              getOrElse(Seq())
```
wajda: Use `Nil` instead of `Seq()`. It's effectively the same but more readable.

core/src/main/scala/za/co/absa/spline/core/harvester/DataLineageBuilder.scala
```diff
@@ -62,11 +75,25 @@ class DataLineageBuilder(logicalPlan: LogicalPlan, executedPlanOpt: Option[Spark
...
+
+            val jdbc = jdbcCommandParser.
+              asWriteCommandIfPossible(curOpNode).
+              map(wc => Seq(wc.query)).
+              getOrElse(Seq())
+
+            var newNodesToProcess: Seq[LogicalPlan] = writes ++ tables ++ jdbc
```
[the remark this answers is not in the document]
wajda: +1 for avoiding `var`s and re-assignments.

core-spark-adapter-api/src/main/scala/za/co/absa/spline/coresparkadapterapi/WriteCommandParser.scala
```diff
@@ -23,18 +23,29 @@ import scala.language.implicitConversions
...
-    tag.runtimeClass.isAssignableFrom(operation.getClass)
-  }
+  def matches(operation: LogicalPlan): Boolean
 
-  def asWriteCommand(operation: T): WriteCommand
+  def asWriteCommand(operation: T): AbstractWriteCommand
 
-  def asWriteCommandIfPossible(operation: T): Option[WriteCommand] =
+  def asWriteCommandIfPossible(operation: T): Option[AbstractWriteCommand] =
     if (matches(operation)) Some(asWriteCommand(operation))
```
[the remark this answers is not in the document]
lukasnalezenec: I don't want to do refactoring when working on my first task on the project. I took already existing code.

core/src/main/scala/za/co/absa/spline/core/harvester/operationBuilders.scala
```diff
@@ -124,6 +124,52 @@ abstract class WriteNodeBuilder
     writeMetrics = writeMetrics,
     readMetrics = readMetrics
   )
+
+  override def ignoreLineageWrite:Boolean = {
+    writeMetrics.get("numFiles").filter(0.>).isDefined
```
wajda: why < 0? Is it correct?
When write was ignored the `numFiles` are set to explicit 0, not a negative value.

core-spark-adapter-2.2/src/main/scala/za/co/absa/spline/coresparkadapterapi/WriteCommandParserImpl.scala
```diff
@@ -16,10 +16,37 @@
...
+
+class SaveAsTableCommandParserImpl extends WriteCommandParser[LogicalPlan] {
+  override def matches(operation: LogicalPlan): Boolean = operation.isInstanceOf[CreateDataSourceTableAsSelectCommand]
+
+  override def asWriteCommand(operation: LogicalPlan): AbstractWriteCommand = {
+    val op = operation.asInstanceOf[CreateDataSourceTableAsSelectCommand]
+
+    val identifier = op.table.storage.locationUri match {
+      case Some(location) => location.toURL.toString
+      case _ => "table:/" + op.table.identifier.identifier
```
wajda: Why database name isn't used in the identifier?

wajda: If `table:/` is a predefined prefix for any `SaveAsTablerCommand` then it should probably be a encapsulated into the command.

model/src/main/scala/za/co/absa/spline/model/DataLineage.scala
```diff
@@ -42,7 +42,8 @@ case class DataLineage
   operations: Seq[Operation],
   datasets: Seq[MetaDataset],
   attributes: Seq[Attribute],
-  dataTypes: Seq[DataType]
+  dataTypes: Seq[DataType],
+  writeIgnored: Boolean = false
```
wajda: `writeIgnore` shouldn't be a property of data lineage object. What was wrong with the detecting the fact of ignored result from the metrics?

lukasnalezenec: We cannot use metrics for detection for saveAsTable operation. It looks like having this property in DataLineage makes sense.

core-spark-adapter-2.2/src/main/scala/za/co/absa/spline/coresparkadapterapi/WriteCommandParserImpl.scala
lukasnalezenec: database name is part of identifier (if database name is not default)

core/src/main/scala/za/co/absa/spline/core/harvester/operationBuilders.scala
lukasnalezenec: Looks like bug :(

core-spark-adapter-2.4/src/test/scala/za/co/absa/spline/coresparkadapterapi/WriteCommandParserImplSpec.scala
```diff
@@ -34,9 +37,33 @@ class WriteCommandParserImplSpec extends FunSpec with BeforeAndAfterEach with Ma
...
+      when(tableIdentifier.identifier).thenReturn("tableIdentifier")
+
+      val tableMock = mock[CatalogTable]
+      when(tableMock.identifier).thenReturn(tableIdentifier)
+
+      val command = mock[CreateDataSourceTableAsSelectCommand]
+      when(command.table).thenReturn(tableMock)
+      when(command.mode).thenReturn(SaveMode.Append)
+
+      val query = mock[LogicalPlan]
```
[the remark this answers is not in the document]
leborgneromain: Agree

core-spark-adapter-2.4/src/test/scala/za/co/absa/spline/coresparkadapterapi/WriteCommandParserImplSpec.scala
```diff
@@ -34,9 +37,36 @@ class WriteCommandParserImplSpec extends FunSpec with BeforeAndAfterEach with Ma
...
+      when(command.table).thenReturn(mockTable)
+      when(command.mode).thenReturn(SaveMode.Append)
+
+      val query = mock[LogicalPlan]
+      when(command.query).thenReturn(query)
+      val factory = new WriteCommandParserFactoryImpl()
+      val instance = factory.getSaveAsTableParser()
+      instance.matches(command) shouldBe true
+      val writeCommand = instance.asWriteCommand(command)
+      writeCommand shouldBe SaveAsTableCommand("table:/tableIdentifier", SaveMode.Append, "table", query)
```
vackosar: Shoudn't the first param be URL? Then I would expect two slashes after "table" scheme:
```URI = scheme:[//authority]path[?query][#fragment]```

File: core-spark-adapter-2.4/src/test/scala/za/co/absa/spline/coresparkadapterapi/WriteCommandParserImplSpec.scala (hunk 8 above)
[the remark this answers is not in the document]
Comment (vackosar) [reply core-spark-adapter-2.4/src/test/scala/za/co/absa/spline/coresparkadapterapi/WriteCommandParserImplSpec.scala]: Interesting. But this is as optional nice to have right?

File: core/src/main/scala/za/co/absa/spline/core/harvester/DataLineageBuilder.scala [hunk 10]
```diff
@@ -62,11 +75,20 @@ class DataLineageBuilder(logicalPlan: LogicalPlan, executedPlanOpt: Option[Spark
           if (parentBuilder != null) parentBuilder += curBuilder
 
           if (maybeExistingBuilder.isEmpty) {
-            val newNodesToProcess =
-              writeCommandParser.
+
+            //try to find all possible commands for traversing- save to filesystem, saveAsTable, JDBC
+            val writes = writeCommandParser.
```
vackosar: Can you really have multiple writes in single query?

core/src/main/scala/za/co/absa/spline/core/harvester/DataLineageBuilder.scala
```diff
@@ -62,11 +75,20 @@ class DataLineageBuilder(logicalPlan: LogicalPlan, executedPlanOpt: Option[Spark
...
                 asWriteCommandIfPossible(curOpNode).
                 map(wc => Seq(wc.query)).
-                getOrElse(curOpNode.children)
+                getOrElse(Nil)
+
+            val tables = tableCommandParser.
```
vackosar: Maybe we could have single WriteNode builder that would find both table writes and file writes. If not then I would rename writeCommand to fileWriteCommand or similar.

core/src/main/scala/za/co/absa/spline/core/harvester/DataLineageBuilder.scala
lukasnalezenec: No, you cannot

core-spark-adapter-2.4/src/test/scala/za/co/absa/spline/coresparkadapterapi/WriteCommandParserImplSpec.scala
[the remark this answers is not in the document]
lukasnalezenec: TLDR: I just copied existing junit test for write operation. This is basically not my code.
I was thinking if this junit test is worth.  And it is. It tests just one line. Its not much.
I would prefer that we should not use any mocks in this test - we should use real classes.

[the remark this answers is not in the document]
wajda: First of all, any code that is copied from anywhere becomes _your_ code and hence is a subject for review as usual.
Second, I'm not sure I understand what you mean by using real classes for tests (speaking about unit tests).

core/src/main/scala/za/co/absa/spline/core/harvester/DataLineageBuilder.scala
vackosar: ok, the comment is bit confusing then maybe even the code could be written more in line with that.

lukasnalezenec: Yes, it is. I wanted to say that it handles all types of commands, but there is only one at time.

PR #149, opened 2019-03-11
Spline #148 add execution info in details panel

leborgneromain:


spline-shadow/client-ui/src/app/components/lineage/lineage-details/schema-details/schema-details.component.html
```diff
@@ -30,8 +30,25 @@
     <schema [schemaId]="'output_schema'" [schemaType]=1 [schema]="getOutput()"></schema>
   </div>
 </div>
-<div *ngIf="!detailsInfo">
-  <div class="details-title">
-    No details to show
+<div *ngIf="!detailsInfo && executedLogicalPlanVM">
```
vackosar: Maybe u could have```*ngIf="!detailsInfo``` on the higher level not having to repeat it.

spline-shadow/client-ui/src/app/services/lineage/lineage-graph.service.ts
```diff
@@ -52,10 +55,12 @@ export class LineageGraphService {
     let that = this
     return that.executionPlanControllerService.lineageUsingGETResponse(executionPlanId).pipe(
       map(response => {
-        return that.convertFromExecutedLogicalPlanToFromExecutedLogicalPlanViewModel(response)
+        let convertedResponse = that.convertFromExecutedLogicalPlanToFromExecutedLogicalPlanViewModel(response)
```
vackosar: convertFromExecutedLogicalPlanTo**From**ExecutedLogicalPlanViewModel
Extra "from"?
Could simplify also as "toLogicalPlanView" rest could be seen from function input n output types.

spline-shadow/client-ui/src/app/services/lineage/lineage-graph.service.ts
```diff
@@ -52,10 +55,12 @@ export class LineageGraphService {
     let that = this
     return that.executionPlanControllerService.lineageUsingGETResponse(executionPlanId).pipe(
```
vackosar: Does returning from constructor do something?

spline-shadow/client-ui/src/app/components/lineage/lineage-details/schema-details/schema-details.component.ts
```diff
@@ -31,10 +32,12 @@ export class SchemaDetailsComponent implements AfterViewInit {
 
   public detailsInfo: any = null
```
vackosar: Maybe there could be some interface that could limit what properties ```details``` could have.

spline-shadow/client-ui/src/app/services/lineage/lineage-graph.service.ts
vackosar: This seems like it can be simplified.

spline-shadow/client-ui/src/app/services/lineage/lineage-graph.service.ts
leborgneromain: yes. I will modify it

spline-shadow/client-ui/src/app/components/lineage/lineage-details/schema-details/schema-details.component.html
leborgneromain: Yes we could but I've done that keeping in mind that it would be easier the day we want to put every single panel as an angular lib

spline-shadow/client-ui/src/app/components/lineage/lineage-details/lineage-details.component.html
```diff
@@ -15,10 +15,10 @@
   -->
 
 <div class="details-container">
-  <div class="container-fluid schema-details">
+  <div class="container-fluid" [ngClass]="detailsInfo? 'schema-details' : 'execution-details'">
```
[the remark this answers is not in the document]
leborgneromain: This is just to show execution details if there is no operation node selected. Imo, this panel is responsible to show details to the user. It is not needed to create a new component for showing execution details.

spline-shadow/client-ui/src/app/components/lineage/lineage-details/schema-details/schema-details.component.ts
leborgneromain: yes indeed but I will modify the type when the rest api will be ready and the types defined and exported via swagger. From now, we have nothing defined yet.

spline-shadow/client-ui/src/app/services/lineage/lineage-graph.service.ts
leborgneromain: not really, this is the way to write call to services in angular using rxjs lib.

spline-shadow/client-ui/src/app/services/lineage/lineage-graph.service.ts
```diff
@@ -52,10 +55,12 @@ export class LineageGraphService {
     let that = this
```
wajda: Is it a getter or modifier? The method is called `get...` but in fact it changes state of another observable.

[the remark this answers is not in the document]
leborgneromain: none, stayed from previous code I guess. I've remove it

spline-shadow/client-ui/src/app/components/lineage/lineage-details/schema-details/schema-details.component.ts
```diff
@@ -60,7 +64,11 @@ export class SchemaDetailsComponent implements AfterViewInit {
             instance.expressionType = type
             this.changedetectorRef.detectChanges()
           }
-        });
+        })
+      } else {
+        this.lineageGraphService.executedLogicalPlan.subscribe(executedLogicalPlan => {
```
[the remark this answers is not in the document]
leborgneromain: you are right, refactored.

spline-shadow/client-ui/src/app/services/lineage/lineage-graph.service.ts
```diff
@@ -52,10 +55,12 @@ export class LineageGraphService {
...
     return that.executionPlanControllerService.lineageUsingGETResponse(executionPlanId).pipe(
       map(response => {
-        return that.convertFromExecutedLogicalPlanToFromExecutedLogicalPlanViewModel(response)
+        let convertedResponse = that.convertFromExecutedLogicalPlanToFromExecutedLogicalPlanViewModel(response)
+        this.executedLogicalPlanSubject.next(convertedResponse)
```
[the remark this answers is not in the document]
leborgneromain: I am not sure to understand. I use a pipe here.

spline-shadow/client-ui/src/app/services/lineage/lineage-graph.service.ts
leborgneromain: yes but it gets the latest selected and convert it. It is a getter that does a conversion if you want.

spline-shadow/client-ui/src/app/services/lineage/lineage-graph.service.ts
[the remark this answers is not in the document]
wajda: I meant mapping instead of piping, sorry.
From the context of the given method and its usage it seems that `executedLogicalPlanSubject` is just a stream of `activeRoute.params` flat mapped on the `executionPlanControllerService.lineageUsingGETResponse` via `toLogicalPlanView` converter function.
In other words, you take a stream of route params, then for each value you obtain a stream of lineages (or execution plans), flat them, and the resulted stream is basically your `executedLogicalPlanSubject `.
Doesn't it seem as a one liner? And so you don't even need the entire `getExecutedLogicalPlan` method?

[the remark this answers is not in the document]
leborgneromain: No Alex, the `executedLogicalPlanSubject `is not a stream of `activeRoute.params`. I've done it exactly for the same reason as `detailsInfo `: to pass the data easily to other components, but in fact this `executedLogicalPlanSubject `is useless. I just take the data of the `lineageUsingGETResponse` method and put it to another observable : `executedLogicalPlanSubject `. This one should be remove indeed. Regarding the code again there is another antipatern that I forgot the correct. The component that use the method `getExecutedLogicalPlan ` is using nested subsumption. I will commit a fix for this and remove the `executedLogicalPlanSubject `

PR #154, opened 2019-03-13
Migrate datatypes properly #152

vackosar:


spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
```diff
@@ -17,15 +16,75 @@
 
 package za.co.absa.spline.persistence.model
 
-case class Progress(timestamp: Long, readCount: Long, _key: Option[String] = None, _id: Option[String] = None, _rev: Option[String] = None) extends DocumentOption
+import DataType._
+
+case class Progress(timestamp: Long, readCount: Long,
+  key: Option[String] = None, _id: Option[String] = None, _rev: Option[String] = None) extends DocumentOption
```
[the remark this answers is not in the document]
vackosar: _type is a type hint,

> Not sure I understand the purpose of designing `DataType` hierarchy that way.
> In your hierarchy you use the same three concrete data type subclasses as in the original model (`SimpleType`, `ArrayType` and `StructType`), but their properties are now mixed up.
> 
>     * What is the benefit of making `name` to be a common property of all three types, while it only has a meaning  for `SimpleType` and for the other two it simply duplicates the `_type` property?
> 
>     * Why does `ArrayType` being a homogenous data structure contains a sequence of children types? What are those children?
> 
>     * Why to decompose a sequence of `StructType`'s fields (where a field has a name and type) into two highly coupled sequences, one with only names and the other with only respected types, having them relying on the sort order to maintain the dependency between the two?

- _type is a type hint for deserialization frameworks. It comes from PolymorphicDocumentOption. 
- name is a field I would expect to be used on the frontend or in queries. Since simple types dont contain "Struct" and "Array" there wont't be any clash.
- ArrayType has an array such that it has similar interface to StructType akin to original class which has a method Array#childDataTypeIds. This symmetry should make implementation of AQL queries and frontend simpler.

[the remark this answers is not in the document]
wajda: How will it make it simpler? Can you give an example of AQL query or Angular service where such design will be helpful?

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
```diff
@@ -17,66 +16,116 @@
...
 package za.co.absa.spline.persistence.model
 
-case class Progress(timestamp: Long, readCount: Long, _key: Option[String] = None, _id: Option[String] = None, _rev: Option[String] = None) extends DocumentOption
+case class Progress(timestamp: Long, readCount: Long,
+  _key: Option[String] = None, _id: Option[String] = None, _rev: Option[String] = None) extends DocumentOption
 
-case class Execution(id: String, dataTypes: Seq[DataType], startTime: Option[Long], endTime: Option[Long], extra: Map[String, Any], _key: Option[String] = None, _id: Option[String] = None, _rev: Option[String] = None) extends DocumentOption
+case class Execution(id: String, dataTypes: Seq[DataType], startTime: Option[Long], endTime: Option[Long],
+  extra: Map[String, Any], _key: Option[String] = None, _id: Option[String] = None, _rev: Option[String] = None)
+  extends DocumentOption
```
wajda: What is `DocumentOption`?

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
```diff
@@ -17,66 +16,116 @@
...
-case class DataType(id: String, name: String, nullable: Boolean, childrenIds: Seq[String])
+trait DataType {
+  def id: String
+  def nullable: Boolean
+}
 
-case class Schema(attributes: Seq[Attribute])
+case class SimpleDataType(
+    id: String,
+    nullable: Boolean,
```
wajda: Let's adopt the rule to always use `override` keyword when we implement or override something, even on attributes. Although Scala compiler doesn't require it I found it very helpful during refactoring and code reading.

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
vackosar: It is similarly to PolymorphicDocumentOption an trait provided by Scarango and was as well added into our code in serialization PR: https://github.com/AbsaOSS/spline/pull/130/files .
There is another one called "Edge". 

In Scarango these were also used to require essential params to be defined on all documents and all edges correspondingly. I think it would be useful for that same purpose for us even now. So I will add those into them.

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
vackosar: added override

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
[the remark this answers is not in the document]
vackosar: Intent was to have united schema e.g. there is method in our original code base ```spline.model.dt.Array#childDataTypeIds``` which returns an array on single value. So you dont have to write if-statements when looking for childDatatypeIds.

But the other option has also its advantages as well. So as agreed I went with the alternative approach.

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
```diff
@@ -17,66 +16,201 @@
...
+  ) extends ArangoDocument
+
+trait DataType {
+  def id: String
+  def nullable: Boolean
+}
 
-case class Execution(id: String, dataTypes: Seq[DataType], startTime: Option[Long], endTime: Option[Long], extra: Map[String, Any], _key: Option[String] = None, _id: Option[String] = None, _rev: Option[String] = None) extends DocumentOption
+case class SimpleDataType(
+    override val id: String,
```
leborgneromain: Can it be _id ? to keep consistency ?

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
```diff
@@ -17,66 +16,201 @@
...
+    dataTypeId: String
+  ) {
 
-trait DocumentOption
+  def this() = this("", "")
+}
 
-trait Edge
+case class StructDataType(
+    override val id: String,
```
leborgneromain: Can it be _id ? to keep consistency ?

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
```diff
@@ -17,66 +16,201 @@
...
 
-trait Operation extends PolymorphicDocumentOption {
-  def name: String
+  def this() = this("", true, null)
 
-  def expression: String
+}
+
+case class ArrayDataType(
+    override val id: String,
```
leborgneromain: Can it be _id ? to keep consistency ?

spline-shadow/persistence/src/main/scala/za/co/absa/spline/persistence/model/persistenceModel.scala
leborgneromain: And String works of course but I think it can be UUID

Issue #175, opened 2019-03-31
spline-spark-adapter-api has spark dependencies in compile scope


igreenfield:
in the pom there are dependencies to spark so it create mixed spark version...

wajda:
What do you mean by mixed spark version? Where is it created?

wajda:
Ok, I think I got the point.
Spark should be defined as "provided", for sure.

igreenfield:
When this is going to be in release version?

PR #176, opened 2019-04-01
Spline #175 Change dependencies of spark to the provided scope 

leborgneromain:
+cleanup unused dependencies

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -50,6 +50,10 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.scalaz</groupId>
+            <artifactId>scalaz-core_${scala.compat.version}</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.spark</groupId>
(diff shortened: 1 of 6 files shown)


vackosar:
good spot!

core/pom.xml
```diff
@@ -66,6 +58,16 @@
             <groupId>org.scalaz</groupId>
             <artifactId>scalaz-core_${scala.compat.version}</artifactId>
         </dependency>
+        <dependency>
```
vackosar: Do you need these at all? Aren't they inherited from spline-spark-api?

leborgneromain: yes for the sample project at least

vackosar: What I mean is that ```spline-core``` inherits ```spark-core_${scala.compat.version}``` and ```spark-sql_${scala.compat.version}``` from ```spline-core-spark-adapter-api```. So you shouldn't need to specify them again in ```spline-core```.

vackosar: sry actually provided scope is not transitive

Issue #166, opened 2019-03-26
Spline UI: search is not working


arvindshmicrosoft:
I am running the UI app JAR locally and while the main functions seem to work correctly, the search function seems to be broken. In that typing in a search string is not returning anything (i.e. the left pane gets blank). I debugged that the REST API call (sample below) is returning a blank array [].

http://localhost:8080/rest/dataset/descriptors?q=mysearchstring&asAtTime=1553556446120&offset=0&size=20

There is no firewall or such - the UI is running local to the browser. From the UI output, there is no exception shown. The only variable is that I am using [Azure Cosmos DB Mongo API](https://docs.microsoft.com/en-us/azure/cosmos-db/mongodb-introduction) but again on the server side there is no apparent issue. Is there any further logging I can turn on at the Java side to debug this issue further?

wajda:
What is _mysearchstring_?
The search text occurrences are only looked up in the following attributes:
- sparkContext.applicationId
- sparkContext.appName
- destination file path
- Spline Lineage UUID

algattik:
The issue is due to a compatibility problem in the Cosmos DB MongoDB emulation.
Running an aggregation query against Cosmos DB:

<img width="389" alt="image" src="https://user-images.githubusercontent.com/22099722/55027340-56090100-5005-11e9-834b-0cb9c4447ae7.png">

while:

<img width="427" alt="image" src="https://user-images.githubusercontent.com/22099722/55027268-29ed8000-5005-11e9-9cab-5241ee4923da.png">

In fact the Spline unit tests fail when run against Cosmos DB.

arvindshmicrosoft:
Thank you @algattik for your help isolating this! @wajda is there any way we can change Spline to not use the PCRE search expressions to work around the current limitation in the MongoDB API within Azure Cosmos DB? If there is any way we can change Spline (in a fork) to test the workaround, then we will be glad to try that locally. Just let us know what you might suggest.

algattik:
I confirm that by changing in BaselineLineageDAO:
 -       val regexMatchOnFieldsCriteria = Seq("appId", "appName", "rootOperation.path") map (_ $regex quote(text) $options "i")
to
+       val regexMatchOnFieldsCriteria = Seq("appId", "appName", "rootOperation.path") map (_ $regex text $options "i")

The search works against Cosmos DB. 

[...]

PR #182, opened 2019-04-08
Cosmos DB compatible search

algattik:
Escape special characters one by one when searching Mongo DB, rather than using Pattern.quote(), for Cosmos DB compatibility. (Azure Cosmos DB can be configured with a Mongo DB compatible API, but it supports ECMAScript regexes instead of PCRE).

Fixes AbsaOSS/spline#166.

### DIFF
--- a/persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
+++ b/persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
@@ -31,6 +31,7 @@ import za.co.absa.spline.persistence.api.DataLineageReader.{PageRequest, Timesta
 import za.co.absa.spline.persistence.mongo.MongoImplicits._
 import za.co.absa.spline.persistence.mongo.dao.BaselineLineageDAO.Component
 import za.co.absa.spline.persistence.mongo.dao.BaselineLineageDAO.Component.SubComponent
+import za.co.absa.spline.persistence.mongo.dao.BaselineLineageDAO.NON_WORD_CHAR
 import za.co.absa.spline.persistence.mongo.{DBCursorToCloseableIterableAdapter, MongoConnection}
 
 import scala.collection.JavaConverters._
@@ -46,10 +47,6 @@ abstract class BaselineLineageDAO extends VersionedLineageDAO with Logging {
 
   protected lazy val dataLineageCollection: DBCollection = getMongoCollectionForComponent(Component.Root)
   protected lazy val operationCollection: DBCollection = getMongoCollectionForComponent(Component.Operation)
-
-  object BaselineLineageDao {
-    final val NON_WORD_CHAR = "\\W".r
-  }
 
   override def save(lineage: DBObject)(implicit e: ExecutionContext): Future[Unit] = {
     val lineageId = lineage.get(idField).toString
@@ -254,7 +251,7 @@ abstract class BaselineLineageDAO extends VersionedLineageDAO with Logging {
     * @return A regular expression searching for the exact literal text string
     */
   private def quoteSafely(text: String): String = {
-    BaselineLineageDao.NON_WORD_CHAR.replaceAllIn(text, m => s"\\\\${m.matched}")
+    NON_WORD_CHAR.replaceAllIn(text, m => s"\\\\${m.matched}")
   }
 
   /**
@@ -296,6 +293,8 @@ abstract class BaselineLineageDAO extends VersionedLineageDAO with Logging {
 }
 
 object BaselineLineageDAO {
+
+  final val NON_WORD_CHAR = "\\W".r
 
   abstract class Component(val name: String)
 
--- a/persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
+++ b/persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
@@ -47,7 +47,9 @@ abstract class BaselineLineageDAO extends VersionedLineageDAO with Logging {
   protected lazy val dataLineageCollection: DBCollection = getMongoCollectionForComponent(Component.Root)
   protected lazy val operationCollection: DBCollection = getMongoCollectionForComponent(Component.Operation)
 
-  private final val NON_WORD_CHAR = "\\W".r
+  object BaselineLineageDao {
+    final val NON_WORD_CHAR = "\\W".r
+  }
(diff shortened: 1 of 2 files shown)


wajda:
Thank you @algattik, we really appreciate your contribution!
Could you please add a unit test that tests your function in different boundary conditions?

Also, there are a few rather minor things that I'd like to outline in the comments below. It would be cool if you fix them as well while you are on it.

Just for information, we are now working on the Contribution guide applied not only for Spline but for all ABSA OSS projects, that aims to unify coding standards and make the contribution process as smooth as possible for both externals as well as our internal contributors. It will be published shortly.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
```diff
@@ -48,6 +47,8 @@ abstract class BaselineLineageDAO extends VersionedLineageDAO with Logging {
   protected lazy val dataLineageCollection: DBCollection = getMongoCollectionForComponent(Component.Root)
   protected lazy val operationCollection: DBCollection = getMongoCollectionForComponent(Component.Operation)
 
+  private final val NON_WORD_CHAR = "\\W".r
```
wajda: Consider placing constants into a companion object.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
```diff
@@ -235,13 +236,25 @@ abstract class BaselineLineageDAO extends VersionedLineageDAO with Logging {
...
 
+  /**
+    * Escape special characters in a regular expression.
+    * Does not use Pattern.quote() for Cosmos DB compatibility.
+    * @see <a href="https://github.com/AbsaOSS/spline/issues/166">issue #166</a>
+    *
+    * @param text A string
+    * @return A regular expression searching for the exact literal text string
+    */
+  protected def quoteSafely(text: String): String = {
```
wajda: Is there any reason for the method being `protected` instead of `private` ?

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
```diff
@@ -235,13 +236,25 @@ abstract class BaselineLineageDAO extends VersionedLineageDAO with Logging {
...
+    return NON_WORD_CHAR.replaceAllIn(text, "\\\\" + _.matched)
```
wajda: Minor: generally we prefer string interpolation over string concatenation. Also consider using triple-quotes to avoid escaping back-slashes.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
algattik: overengineering probably :) thought it could be an extension point for other storage implementations

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
algattik: Unfortunately triple-quotes don't reduce the number of required backslashes.

algattik:
Thanks a lot @wajda, great feedback! I've improved the code added a unit test.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
```diff
@@ -47,7 +47,9 @@ abstract class BaselineLineageDAO extends VersionedLineageDAO with Logging {
   protected lazy val dataLineageCollection: DBCollection = getMongoCollectionForComponent(Component.Root)
   protected lazy val operationCollection: DBCollection = getMongoCollectionForComponent(Component.Operation)
 
-  private final val NON_WORD_CHAR = "\\W".r
+  object BaselineLineageDao {
```
wajda: No, I meant place it into the companion object of the `BaselineLineageDAO` class. It already exists at the bottom of the file.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
wajda: Why? There is no escape characters in Scala's multi-line string literal.
```scala
scala> "\\\\"
res0: String = \\

scala> """\\\\"""
res1: String = \\\\

scala> "\\\\" == """\\"""
res2: Boolean = true
```

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
algattik: Got it. Fixed.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/BaselineLineageDAO.scala
wajda: Interesting :) Thanks.

wajda:
The method `quoteSafely` seems to fail on the input with backslashes. It needs more tests.
```scala
scala> NON_WORD_CHAR.replaceAllIn("this is a text with backslash \\", m => s"\\\\${m.matched}")
java.lang.IllegalArgumentException: character to be escaped is missing
  at java.util.regex.Matcher.appendReplacement(Matcher.java:809)
  at scala.util.matching.Regex$Replacement.replace(Regex.scala:900)
  at scala.util.matching.Regex$Replacement.replace$(Regex.scala:900)
  at scala.util.matching.Regex$MatchIterator$$anon$4.replace(Regex.scala:878)
  at scala.util.matching.Regex.$anonfun$replaceAllIn$1(Regex.scala:508)
  at scala.collection.Iterator.foreach(Iterator.scala:941)
  at scala.collection.Iterator.foreach$(Iterator.scala:941)
  at scala.collection.AbstractIterator.foreach(Iterator.scala:1429)
  at scala.util.matching.Regex.replaceAllIn(Regex.scala:508)
  ... 28 elided
```

algattik:
Fixed replacement for special chars, thanks

PR #193, opened 2019-04-11
codeless init

vackosar:


core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
```diff
@@ -58,31 +69,56 @@ object SparkLineageInitializer extends Logging {
...
+        .getOption(QUERY_EXECUTION_LISTENERS.key).toSeq
+        .flatMap(s => s.split(",").toSeq)
+        .contains(classOf[CodelessQueryExecutionListener].getCanonicalName)
+      if (!splineConfiguredForCodelessInit || spark.SPARK_VERSION.startsWith("2.2")) {
+        if (splineConfiguredForCodelessInit) {
+          log.warn("Spline lineage tracking is also configured for codeless initialization, but codeless init is supported on Spark 2.3+ and not current version 2.2. Spline will be initialized only via code call to enableLineageTracking i.e. the same way as is now.")
         }
+        createQueryExecutionListener(configurer)
+          .foreach(sparkSession.listenerManager.register(_))
+        // TODO: SL-128
```
wajda: Please remove all TODOs. According to the agreement all TODOs should be either resolved in the current PR or a new issue should be created on GitHub. In exceptional cases, when a TODO cannot be fixed now but really helps readability (or, to be more precise, absence of it would hurt the code readability) then it can be left with the bi-directional reference to the respective github issue.

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
```diff
@@ -33,12 +38,18 @@ import scala.util.control.NonFatal
...
-  }
 
-  def enableLineageTracking(sparkSession: SparkSession, configurer: SplineConfigurer): SparkSession = {
+  def enableLineageTracking(sparkSession: SparkSession, configurer: SplineConfigurer): SparkSession =
     SparkSessionWrapper(sparkSession).enableLineageTracking(configurer)
+
+  def createSplineOrNoopQueryExecutionListener(sparkSession: SparkSession): QueryExecutionListener =
+    SparkSessionWrapper(sparkSession).createSplineOrNoopQueryExecutionListener()
+
+  class NoOpQueryExecutionListener extends QueryExecutionListener {
```
wajda: Why do you need this class? Wouldn't a simple `Option` suffice?

sample/src/main/scala/za/co/absa/spline/sample/batch/CodelessInitSampleJob.scala
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.spark.sql.SaveMode
+import za.co.absa.spline.sample.SparkApp
+
+/**
+  * A sample job where no explicit Spline code was used to initialize Spline. Only Spark configuration was used.
+  */
+object CodelessInitSampleJob extends SparkApp(
+  name ="Codeless Init Sample Job",
+  // Spark configuration used to register Spline listener for codeless init.
+  conf = Seq(("spark.sql.queryExecutionListeners", "za.co.absa.spline.core.listener.CodelessQueryExecutionListener"))) {
```
wajda: I would avoid using the word 'codeless' in the execution listener class name. It sounds confusing. Can we rename it to just `QueryExecutionListener`?

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
vackosar: I need to delegate to actual implementation. It would be possible to have an ```Option#isDefined``` check on each call to CodelessQueryExecutionListener.

sample/src/main/scala/za/co/absa/spline/sample/batch/CodelessInitSampleJob.scala
vackosar: I don't think ```QueryExecutionListener``` is a good idea as it would be duplicating name of ```QueryExecutionListener``` although in different package. At least we should suffix it with ```Impl```.
It would be better to give the class name that identifies its purpose. Here I am assuming that we will call the feature "codeless init", thus the name. I would be possibly even better to call it ```CodelessInitQueryExecutionListener```, so something akin that.

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
```diff
@@ -58,31 +69,56 @@ object SparkLineageInitializer extends Logging {
...
-        } catch {
-          case NonFatal(e) if configurer.splineMode == BEST_EFFORT =>
-            log.error(s"Initialization failed! Spark Lineage tracking is DISABLED.", e)
+      val splineConfiguredForCodelessInit = sparkSession.sparkContext.getConf
+        .getOption(QUERY_EXECUTION_LISTENERS.key).toSeq
+        .flatMap(s => s.split(",").toSeq)
+        .contains(classOf[CodelessQueryExecutionListener].getCanonicalName)
+      if (!splineConfiguredForCodelessInit || spark.SPARK_VERSION.startsWith("2.2")) {
+        if (splineConfiguredForCodelessInit) {
+          log.warn("Spline lineage tracking is also configured for codeless initialization, but codeless init is supported on Spark 2.3+ and not current version 2.2. Spline will be initialized only via code call to enableLineageTracking i.e. the same way as is now.")
```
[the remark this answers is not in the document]
vackosar: Messages in single line have advantage that are easily searched as the line break don't prevent match.

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
wajda: What's wrong with `Option.forEach`?

core/src/test/scala/za/co/absa/spline/core/SparkLineageInitializerSpec.scala
```diff
@@ -125,13 +138,40 @@ class SparkLineageInitializerSpec extends FunSpec with BeforeAndAfterEach with M
     }
   }
 
+  describe("codeless initialization") {
+    it("should not allow duplicate tracking when combining the methods") {
+      jvmProps.setProperty(QUERY_EXECUTION_LISTENERS.key, s"${classOf[CodelessQueryExecutionListener].getName}")
```
lukasnalezenec: nip: unnecessary string interpolation

sample/src/main/scala/za/co/absa/spline/sample/batch/CodelessInitSampleJob.scala
```diff
@@ -0,0 +1,40 @@
...
+
+package za.co.absa.spline.sample.batch
+
+import org.apache.spark.sql.SaveMode
+import za.co.absa.spline.sample.SparkApp
+
+/**
+  * A sample job where no explicit Spline code was used to initialize Spline. Only Spark configuration was used.
+  */
+object CodelessInitSampleJob extends SparkApp(
```
lukasnalezenec: IMHO we should have at least one simple end to end test that will be run in unit test.
It should use just make Spark context, do some computation and check if computation was recorded.
Even smoke test would be sufficient.

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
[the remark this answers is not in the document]
wajda: It is also ugly, unreadable and creates annoying horizontal scrollbar in viewers.
Nobody will search it as a whole thing in the source code.
Please brake it into several lines.

sample/src/main/scala/za/co/absa/spline/sample/batch/CodelessInitSampleJob.scala
vackosar: I wanted add some, but on develop branch only to avoid too much conflicts.

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
```diff
@@ -58,31 +63,59 @@ object SparkLineageInitializer extends Logging {
...
+        None
+      }
+    }
 
-      // TODO: SL-128
-//       sparkSession.streams addListener configurer.streamingQueryListener
+    private def createEventHandler(configurer: SplineConfigurer): Option[QueryExecutionEventHandler] = {
+      if (configurer.splineMode != DISABLED) {
+        if (!getOrSetIsInitialized()) {
+          log info s"Spline v${SplineBuildInfo.version} is initializing..."
```
wajda: `log info ""` versus `log.error("")`
Generally we should avoid using infix notation excepts for DSL, binary operators and cases when it helps readability. And when we do use it we should use it consistently.

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
```diff
@@ -58,31 +63,59 @@ object SparkLineageInitializer extends Logging {
...
+          try {
+            SparkVersionRequirement.instance.requireSupportedVersion()
+            val eventHandler = configurer.queryExecutionEventHandler
+            log info s"Spline successfully initialized. Spark Lineage tracking is ENABLED."
+            Some(eventHandler)
+          } catch {
+            case NonFatal(e) if configurer.splineMode == BEST_EFFORT =>
+              log.error(s"Spline initialization failed! Spark Lineage tracking is DISABLED.", e)
```
wajda: Shouldn't it be warning?

core/src/main/scala/za/co/absa/spline/core/listener/SplineQueryExecutionListener.scala
```diff
@@ -16,55 +16,38 @@
...
-    * The method is executed when an error occurs during an action execution.
-    *
-    * @param funcName  A name of the executed action.
-    * @param qe        A Spark object holding lineage information (logical, optimized, physical plan)
-    * @param exception An exception describing the reason of the error
-    */
-  def onFailure(funcName: String, qe: QueryExecution, exception: Exception): Unit = {
-    log.error(s"Action '$funcName' execution failed", exception)
+  private def constructEventHandler(): Option[QueryExecutionEventHandler] = {
+    val sparkSession = SparkSession.getActiveSession.getOrElse(
```
wajda: Two nested `getOrElse` can be flattened and chained by using `orElse`. E.g.
```scala
 SparkSession.getActiveSession
  .orElse(SparkSession.getDefaultSession)
  .getOrElse(throw new IllegalStateException(...))
```

core/src/test/scala/za/co/absa/spline/core/SparkLineageInitializerSpec.scala
```diff
@@ -47,25 +49,35 @@ object SparkLineageInitializerSpec {
...
         (f get session.listenerManager).asInstanceOf[Seq[QueryExecutionListener]].map(_.getClass)
     }).get
   }
 
-  private def assertSplineIsEnabled() = sparkQueryExecutionListenerClasses should contain(classOf[SplineQueryExecutionListener])
+  private def assertSplineIsEnabled(session: SparkSession = SparkSession.builder().getOrCreate()) =
+    sparkQueryExecutionListenerClasses(session) should contain(classOf[SplineQueryExecutionListener])
+
+  private def assertSplineIsDisabled(session: SparkSession = SparkSession.builder().getOrCreate()) =
+    sparkQueryExecutionListenerClasses(session) should not contain classOf[SplineQueryExecutionListener]
```
wajda: nit: inconsistent braces around `classOf`

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
vackosar: I personally don't like infix, but this is another part of previous code. I don't think it is good idea to try to fix all issues with previously written code when ever we touch it.

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
vackosar: In this case I would use error. Initialization failed due to unhandled exception, which we here consume. The Spline code cannot continue, so I would raise it on error log level, such that it is not easily ignored.

core/src/test/scala/za/co/absa/spline/core/SparkLineageInitializerSpec.scala
vackosar: also previous code issue.

core/src/main/scala/za/co/absa/spline/core/SparkLineageInitializer.scala
vackosar: I removed this, but it also comes from previous code.

core/src/test/scala/za/co/absa/spline/core/SparkLineageInitializerSpec.scala
vackosar: Interesting that the `not` operator causes that if you try to add brackets for the second contain the IDEA will raise warning that the brackets are not necessary. The first brackets cannot be removed. So I am keeping as is.

Issue #131, opened 2019-02-20
Codeless Spline initialization


wajda:
... via `spark.sql.queryExecutionListeners` Spark config property

wajda:
This might also solve the following issues:
- #48 
- #43 
- #106

wajda:
I think we need to support three way of Spline inialization:
1. Scala mixin - `sparkSession.enableLineageTracking()`
1. Utility method - `SparkLineageInitializer.enableLineageTracking(sparkSession)`
1. Codeless via Spark config

So the users would be able to use whatever they like depending on their technology stack and deployment requirements.

lokm01:
Great idea!

wajda:
https://github.com/apache/spark/commit/7703b46d2843db99e28110c4c7ccf60934412504

PR #199, opened 2019-04-23
CodelessInit for Dev branch #131

vackosar:


integration-tests/src/test/scala/za/co/absa/spline/CodelessInitSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+import org.spark_project.jetty.server.Server
+import org.spark_project.jetty.servlet.ServletHandler
+import za.co.absa.spline.CountingServlet._
+import za.co.absa.spline.harvester.LineageDispatcher.publishUrlProperty
+import za.co.absa.spline.harvester.SparkLineageInitializer.sparkQueryExecutionListenersKey
+import za.co.absa.spline.harvester.listener.SplineQueryExecutionListener
+
+import scala.util.Random
+
+class CodelessInitSpec extends FlatSpec with Matchers with BeforeAndAfterEach with OneInstancePerTest {
```
wajda: Name `CodelessInitSpec` suggests it's all about codeless initialization stuff. Why should we care about REST, RDD, Schema and the other unrelated things?
Can't it be a normal unit test?

integration-tests/src/test/scala/za/co/absa/spline/CodelessInitSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+
+  override def beforeEach(): Unit = {
+    super.beforeEach()
+    resetSparkContext()
```
wajda: I'd say this needs to be a fixture. Maybe just a loan-method in a `SparkFixure`

integration-tests/src/test/scala/za/co/absa/spline/CodelessInitSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+  /**
+    * SparkContext needs to be cleared before and after to prevent propagation of test-specific SparkConf.
+    */
+  private def resetSparkContext(): Unit = {
+    SparkSession.getActiveSession.orElse(SparkSession.getDefaultSession).foreach(_.stop())
+  }
+
+}
+
+class CountingServlet extends HttpServlet {
```
wajda: `CodelessInitSpec.scala` file is supposed to only contain `CodelessInitSpec` class and its companion object.
`CountingServlet` class should be placed into its own file.

integration-tests/src/test/scala/za/co/absa/spline/CodelessInitSpec.scala
vackosar: For this end2end test to be converted to unit test we would need to provide some way to mock LineageDispatcher via configuration properties e.g. something like what we had for persistence factories.

integration-tests/src/test/scala/za/co/absa/spline/CodelessInitSpec.scala
vackosar: I was afraid that it would encourage resetting Spark context, which could impact other concurrently running tests. I hope this would be only needed in this specific codeless init test.

harvester/src/test/scala/za/co/absa/spline/harvester/SparkLineageInitializerSpec.scala
```diff
@@ -16,60 +16,68 @@
...
+import scala.collection.JavaConverters._
 
 object SparkLineageInitializerSpec {
 
-  class MockReadWritePersistenceFactory(conf: Configuration) extends PersistenceFactory(conf) with MockitoSugar {
-    override val createDataLineageWriter: DataLineageWriter = mock[DataLineageWriter]
-    override val createDataLineageReader: Option[DataLineageReader] = Some(mock[DataLineageReader])
-    override val createProgressEventWriter: ProgressEventWriter = mock[ProgressEventWriter]
+  private[this] def getSparkQueryExecutionListenerClasses(session: SparkSession): Seq[Class[_ <: QueryExecutionListener]] = {
+    val listeners: Seq[QueryExecutionListener] = getFieldValue(session.listenerManager.clone(),
```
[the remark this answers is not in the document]
vackosar: Do we follow some specific style guide? Perhaps something published on the net?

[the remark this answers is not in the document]
wajda: https://docs.scala-lang.org/style/indentation.html#methods-with-numerous-arguments

PR #201, opened 2019-04-24
spline-188 Exception "requires value for 'writeMetrics'

lukasnalezenec:


persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/LineageDAOv4.scala
```diff
@@ -100,6 +100,15 @@ class LineageDAOv4(override val connection: MongoConnection) extends BaselineLin
             addAll(opTransformations.asJava)
           })
         }
+        else if (isWriteOperation(op)) {
+          if (!op.containsField("writeMetrics")) {
```
vackosar: In operations_v4 some writes have metrics and some don't?

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/LineageDAOv4.scala
```diff
@@ -100,6 +100,15 @@ class LineageDAOv4(override val connection: MongoConnection) extends BaselineLin
...
+            op.put("writeMetrics", new BasicDBObject(new ju.HashMap()))
```
vackosar: Do you need to add something like this also into MutableLineageUpgraderV4#upgrader?

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/LineageDAOv4.scala
wajda: Correct. Metrics were added to the model without changing the model version. The assumption was that since metrics are optional absence of a field would be treated as an empty field (empty map). But in fact Salat doesn't treat is like that. So the purpose of this line of code is to force that default.

persistence/mongo/src/main/scala/za/co/absa/spline/persistence/mongo/dao/LineageDAOv4.scala
wajda: no, because this needs to be executed on v4 objects. The v4Upgrader is called on v3 objects (to upgrade ones to v4)

PR #203, opened 2019-04-26
Bugfix/spline 189 slow ui

wajda:


web/ui/src/app/dataset/lineage-overview/lienage-overview-graph.component.ts
```diff
@@ -80,6 +80,9 @@ export class LineageOverviewGraphComponent implements OnInit, OnDestroy {
 
     ngOnDestroy(): void {
         this.subscriptions.forEach(s => s.unsubscribe())
+        this.network.destroy()
+        delete this.network
```
leborgneromain: is it necessary after calling destroy ?

wajda: It isn't, as far as I understand. But that's what Vis explicitly suggests. No harm no gain, but at least it makes the purpose more explicit.

PR #220, opened 2019-05-06
spline-219 Fix Spark 2.2 build

lukasnalezenec:


core/src/test/scala/za/co/absa/spline/core/SparkLineageInitializerSpec.scala
```diff
@@ -139,13 +138,20 @@ class SparkLineageInitializerSpec extends FunSpec with BeforeAndAfterEach with M
...
   describe("codeless initialization") {
     it("should not allow duplicate tracking when combining the methods") {
-      jvmProps.setProperty(QUERY_EXECUTION_LISTENERS.key, classOf[SplineQueryExecutionListener].getName)
+      jvmProps.setProperty(sparkQueryExecutionListenersKey, classOf[SplineQueryExecutionListener].getName)
       jvmProps.setProperty(PERSISTENCE_FACTORY, classOf[MockWriteOnlyPersistenceFactory].getName)
       val session = SparkSession.builder.getOrCreate
-      numberOfRegisteredSplineListeners(session) shouldBe 1
+      val isSpark22 = session.version.startsWith("2.2")
+
+      if (isSpark22) {
```
wajda: I'll approve it for now, just to speed up the merging process. But it should definitely be replaced with a conditional execution combined with `SparkFixure`. That way you won't have much code duplication and you won't have branching inside you tests. #226

Issue #229, opened 2019-05-09
[mobx] Encountered an uncaught exception that was thrown by a reaction or observer component


wajda:
The bug is apparently harmless, but is worth further investigation.
The following error message appears in browser console when switching from a sample job 3 lineage to another job lineage.
Excepts for the message the UI functionality doesn't seem to be affected. 

```
mobx.module.js:3354 [mobx] Encountered an uncaught exception that was thrown by a reaction or observer component, in: 'Reaction[t.detectChanges()] TypeError: Cannot read property '_typeHint' of undefined
    at o (types.ts:65)
    at Object.t.typeOfDataType (types.ts:61)
    at e.typeOfDataType (data-type-view.component.ts:35)
    at Object.eval [as updateDirectives] (VM28 e.ngfactory.js:59)
    at Object.updateDirectives (core.js:23654)
    at CM (core.js:23307)
    at DM (core.js:23548)
    at mM (core.js:23511)
    at CM (core.js:23308)
    at DM (core.js:23548)
```

PR #231, opened 2019-05-09
spline #229 - [mobx] Encountered an uncaught exception that was throw…

wajda:
…n by a reaction or observer component


wajda:
@leborgneromain

Issue #241, opened 2019-05-22
Fix logging configuration in unit test on develop branch


lukasnalezenec:
Logging in unit test is verbose and it makes working with tests very hard

PR #243, opened 2019-05-22
spline #241 logging in unit test.

lukasnalezenec:


parent/pom.xml
```diff
@@ -555,21 +554,21 @@
...
+<!--            <dependency>-->
+<!--                <groupId>ch.qos.logback</groupId>-->
+<!--                <artifactId>logback-classic</artifactId>-->
+<!--                <version>${logback.version}</version>-->
+<!--            </dependency>-->
+<!--            <dependency>-->
+<!--                <groupId>ch.qos.logback</groupId>-->
+<!--                <artifactId>logback-access</artifactId>-->
+<!--                <version>${logback.version}</version>-->
+<!--            </dependency>-->
```
leborgneromain: Why not delete these dependencies completely instead of commenting them if they are not used ?

Issue #244, opened 2019-05-22
fix build on Develop for Spark 2.2


lukasnalezenec:
When running Spline with Spark 2.2, it tries to access fields that does not exist in Spark 2.2

 should warn on double initialization *** FAILED ***
  java.lang.NoSuchFieldException: org$apache$spark$util$ListenerBus$$listenersPlusTimers
  at java.lang.Class.getDeclaredField(Class.java:2070)
  at za.co.absa.spline.common.ReflectionUtils$.getFieldValue(ReflectionUtils.scala:51)
  at za.co.absa.spline.harvester.SparkLineageInitializerSpec$.getStreamingListenerClasses(SparkLineageInitializerSpec.scala:55)
  at za.co.absa.spline.harvester.SparkLineageInitializerSpec$.numberOfRegisteredStreamListeners(SparkLineageInitializerSpec.scala:68)
  at za.co.absa.spline.harvester.SparkLineageInitializerSpec$$anonfun$3$$anonfun$apply$mcV$sp$5$$anonfun$apply$9.apply(SparkLineageInitializerSpec.scala:175)
  at za.co.absa.spline.harvester.SparkLineageInitializerSpec$$anonfun$3$$anonfun$apply$mcV$sp$5$$anonfun$apply$9.apply(SparkLineageInitializerSpec.scala:172)
  at za.co.absa.spline.fixture.SparkFixture$class.withNewSparkSession(SparkFixture.scala:39)
  at za.co.absa.spline.harvester.SparkLineageInitializerSpec.withNewSparkSession(SparkLineageInitializerSpec.scala:73)
  at za.co.absa.spline.harvester.SparkLineageInitializerSpec$$anonfun$3$$anonfun$apply$mcV$sp$5.apply(SparkLineageInitializerSpec.scala:172)
  at za.co.absa.spline.harvester.SparkLineageInitializerSpec$$anonfun$3$$anonfun$apply$mcV$sp$5.apply(SparkLineageInitializerSpec.scala:172)

PR #245, opened 2019-05-22
spline #244 Ignore test by Spark version

lukasnalezenec:


test-commons/src/main/scala/za/co/absa/spline/scalatest/ConditionalIgnore.scala
```diff
@@ -0,0 +1,15 @@
+package za.co.absa.spline.scalatest
```
leborgneromain: missing the ©

lukasnalezenec: Thanks, fixed

harvester/src/test/scala/za/co/absa/spline/harvester/SparkLineageInitializerSpec.scala
```diff
@@ -90,7 +91,7 @@ class SparkLineageInitializerSpec extends FunSpec with BeforeAndAfterEach with M
     val valueFromJVM = "value from JVM args"
     val valueFromSpline = "value from spline.properties"
 
-    it("should look through the multiple sources for the configuration properties") {
+      it("should look through the multiple sources for the configuration properties") {
```
wajda: nit: indent

harvester/src/test/scala/za/co/absa/spline/harvester/SparkLineageInitializerSpec.scala
```diff
@@ -139,7 +140,7 @@ class SparkLineageInitializerSpec extends FunSpec with BeforeAndAfterEach with M
   }
 
   describe("codeless initialization") {
-    it("should not allow duplicate tracking when combining the methods") {
+     it("should not allow duplicate tracking when combining the methods", IGNORE_SPARK_22) {
```
wajda: nit: indent

test-commons/src/main/scala/za/co/absa/spline/scalatest/ConditionalIgnore.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package za.co.absa.spline.scalatest
+
+import org.scalatest.{Ignore, Tag}
+import org.apache.spark.SPARK_VERSION
+
+object ConditionalTestIgnore {
+
+  private def ignoreTestBySparkVersion(condition: String => Boolean): Tag = {
```
wajda: The method name "ignore..." implies that something should be ignored when the `condition` holds. The implementation does the opposite thing - returns `@Ignore` tag when `condition` doesn't hold.

Also why does the `condition` function take a Spark version argument, considering the fact that one is a constant? Can't it be inlined?

test-commons/src/main/scala/za/co/absa/spline/scalatest/ConditionalIgnore.scala
```diff
@@ -0,0 +1,31 @@
...
+    val matches = condition(SPARK_VERSION)
+    Tag(if (matches) "" else classOf[Ignore].getName)
+  }
+
+  //Ignores unit test when version of spark is 2.2 or below.
```
wajda: This comment is both misleading (should be exactly 2.2) and redundant. The line of code is a self-explanatory.

test-commons/src/main/scala/za/co/absa/spline/scalatest/ConditionalIgnore.scala
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.scalatest
+
+import org.scalatest.{Ignore, Tag}
+import org.apache.spark.SPARK_VERSION
+
+object ConditionalTestIgnore {
```
wajda: The object name doesn't match the file name.

test-commons/src/main/scala/za/co/absa/spline/scalatest/ConditionalIgnore.scala
lukasnalezenec: 1. I was not able to find good opposite from "ignore"

wajda: what if we simply rename it to, say `ignoreUnless(...)`, so that the name fits the signature?
Also what about making independent from the Spark version specifically? If you change the condition type to just `condition: => Boolean` you could use it in test directly like this:
```scala
it("should bla-bla-bla", ignoreUnless(!SPARK_VERSION.startsWith("2.2")) {
  ...
}
```
Or even better, to get rid of negations we could use another method name (might be an additional method delegating to the first one) - `ignoreWhen(SPARK_VERSION.startsWith("2.2"))`

PR #253, opened 2019-05-29
Scala 2.12+ cross-compilation

igreenfield:


wajda:
I've released `velocypack1.4-module-scala` for Scala 2.12 - https://github.com/wajda/java-velocypack-module-scala/tree/mvn-repo/com/github/wajda/com/arangodb

wajda:
I don't think there is a need for the following modules to have a Scala version suffix:
- `client-ui` (doesn't contain any bytecode)
- `client-webjar` (doesn't contain any bytecode)
- `client-web`
- `rest-gateway`
- `rest-api-doc-generator`

The first two modules only contain client-side javascript stuff and have nothing to do with Scala.
The last three modules are either WARs or CLI jars and are not intended for reuse. In other words they don't serve as compile time dependencies for anything else - they are final runnable components.

wajda:
Actually, I just thought that maybe we don't need to cross-compile the entire project excepts for only spark-related modules, do we? Will t make it simpler if we only include `spark/*` modules to the `scala-2.12` profile?
One reason for that thinking is that cross-compiling only makes sense for modules that are intended to be reused from outside of the Spline project. Those are mainly spark agent related modules. For the rest of the project it's enough to only have a one selected Scala version and just use it.
The second reason is that by minimizing the variety of Scala versions used throughout the project, we also minimize the chance that in the future we run into a 3rd-party dependency, that wouldn't support scala 2.12 (as in the case with velocypack for example). So we'll minimize potential problems and make the dev life easier.
What do you think?

igreenfield:
@wajda 
1. The Spark way of cross build is all or nothing.
2. I think that if module is most permanent it should be moved to separate repo.

wajda:
I agree, but now we cannot afford splitting the project while it's under active development. We'll do it later when we have more or less stable API between the modules.
But why do say it's all or nothing? If I understand it correctly, the approach you've taken allows us to selectively cross-compile modules. By default all POMs use 2.11 and all modules are included into the build. Then I want to compile certain modules against other Scala versions. So I run your script say with `scala_ver=2.11` and it updates the `artefactId` for certain modules (we can have a hardcoded list of them) and run the Maven build with the respective profile that would only include those affected modules in the the build. We just need to properly maintain the module list, so that it forms a complete dependency tree for those modules. All the other modules that are not in the list (e.g. WAR, CLI etc, that are not intended to be used as a dependencies themselves) will just use 2.11 versions of everything and it will be perfectly enough.

igreenfield:
You will need to change the deploy hence the build produces the same artifacts twice... (these that you do not change... from 2.11 to 2.12)
What do you think?

wajda:
Those artefacts that do not change won't be included into the build. That is the point - include all modules in the default (2.11) build, but only the modules that change into the cross-compile builds.

wajda:
Example:

I have modules A, B, C, D.
Module A - is a WAR file, B - is a JAR that might be used by other external applications a a dependency, and C, D are just some common modules, services and such, that are used by A and B. The dependency trees look like this `A -> C -> D` and `B -> C -> D`.
The module B needs to be cross-compiled. So that the dependency trees expand.
`A_2.11 -> C_2.11 -> D_2.11'
`B_2.11 -> C_2.11 -> D_2.11`
`B_2.12 -> C_2.12 -> D_2.12`
So now it's clear that we need to include all 4 modules into 2.11 builds and only B, C and D into 2.12 builds.

wajda:
... since A is only compiled with 2.11 I can omit the version suffix for A. So instead of `A_2.11` I can call it just `A`. This will serve as a hint that there is no cross-compiled versions for this artefact, so it will help to avoid confusions.

igreenfield:
@wajda 
Are you sure about that: client-webjar (doesn't contain any bytecode) as I see in the code there are class there...

igreenfield:
@wajda Done please review

wajda:
sorry for delay. Unfortunately we don't have a capacity for this at the moment. The more I think about cross compilation the more it seems to me that it's better to wait until we extract Spline Spark agent into a separate project. Because the agent is the only module that actually needs to be cross compiled. Doing it for the entire Spline project seems to be too invasive.

igreenfield:
@wajda Hi Alex,
You are right the that this is the only one that MUST... but staying on 2.11 is not a good idea. hence 2.13 is already released so 2.11 is going to be deprecated. and the migration path will much complicated even now you have APIs that changed from 2.11 to 2.12 as long as you wait with this it will be harder.
Moreover, many Spark projects move to 2.12 as spark moves to there so they won't use spline... and will find other projects to use...
So I think it is better to do it as early as possible.
Just to let you know we want to use spline but without 2.12 release we won't be able to do it so we will use other solutions.

wajda:
I hear what you are saying. Yes, of course we surely will support 2.12 and further Scala versions. It's just the matter of priority for us at the moment. We have limited dev capacity and are fully focused on having functionality done first. Spline 0.4 is major Spline redesign. I bet if Spline supported 2.12, but didn't do what it supposed to do properly, it would hardly be useful for your project anyway :) No worries, we'll return to it shortly after 0.4.0 is released. Basically, just in a few weeks.
Thank you for your contribution and patience!

wajda:
Spline Spark agent module is now extracted into a separate [project](/absaoss/spline-spark-agent), and the cross-compilation will be implemented there via `scala-cross-build-maven-plugin` plugin.
Thus closing the PR.

PR #264, opened 2019-06-03
spline #261 - Producer REST model

wajda:


producer-rest-model/src/main/scala/za/co/absa/spline/producer/rest/model/ExecutionEvent.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.producer.rest.model
+
+case class ExecutionEvent(
+  planId: ExecutionPlanId,
+  timestamp: Long
```
leborgneromain: Should we add the errors later ?

producer-rest-model/src/main/scala/za/co/absa/spline/producer/rest/model/operationModel.scala
```diff
@@ -0,0 +1,48 @@
...
+  override val parentIds: Array[OperationId],
+  override val schema: Option[Schema]
+) extends OperationLike
+
+case class ReadOperation(
+  inputSources: Array[SourceUrl],
+  override val id: OperationId,
+  override val schema: Option[Schema] = None
+) extends OperationLike {
+  override val parentIds: Array[OperationId] = Array.empty
```
leborgneromain: Why this is initialized in Read Operation and not the others ?

wajda: That's terminology confusion. I've renamed it to `childIds'

producer-rest-model/src/main/scala/za/co/absa/spline/producer/rest/model/ExecutionEvent.scala
wajda: added

PR #268, opened 2019-06-06
Commons: add JSON serialization utils

wajda:


lukasnalezenec:
This might cause dependency issues with Spark. We should test it if it works with different Spark versions (2.4 and pre2.4) on classpath.
https://issues.apache.org/jira/browse/SPARK-23043

PR #287, opened 2019-06-21
spline-285 Arango database creator (Spline Admin CLI)

wajda:


admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
```diff
@@ -0,0 +1,73 @@
...
+    val cliParser = new scopt.OptionParser[AdminCLIConfig]("Spline Admin CLI") {
+      head("Spline Admin Command Line Interface", SplineBuildInfo.version)
+
+      help("help").text("prints this usage text")
+
+      (cmd("initdb")
+        action ((_, c) => c.copy(cmd = InitDB()))
+        text "Initialize Spline database."
+        children(
+        opt[Unit]('f', "force")
```
leborgneromain: Should it be Boolean ?

wajda: No, it shouldn't. The type here is a type of a value associated with the given key.
```scala
opt[Boolean]('f') // implies the usage: -f <true|false>
opt[Unit]('f')    // means it's a flag: -f
```

leborgneromain: ah ok. make sence. LGTM then

Issue #273, opened 2019-06-14
Simplify the arango storage model


leborgneromain:

leborgneromain:
Note for the dev : Some entities might not need the _key and only the ones that actually need it can define it as non optional attribute.

wajda:
Because of the storage model discrepancy between two different PRs with the regard to `Schema`, the persister, migrator and the consumerb need to be fixed according the the selected model.

wajda:
> Because of the storage model discrepancy between two different PRs with the regard to `Schema`, the persister, migrator and the consumerb need to be fixed according the the selected model.

this will be fixed in #301

Issue #258, opened 2019-05-31
Model for Schema


wajda:
Create more flexible schema/attributes/types model

PR #293, opened 2019-07-03
schema integration

leborgneromain:
Add schema integration : https://github.com/AbsaOSS/spline/issues/258
Simplify the persistence model : https://github.com/AbsaOSS/spline/issues/273


migrator-tool/src/main/scala/za/co/absa/spline/migrator/DataLineageToExecPlanWithEventConverter.scala
```diff
@@ -89,29 +89,43 @@ class DataLineageToExecPlanWithEventConverter(lineage: DataLineage) {
...
         OperationParams.Name -> opRead.mainProps.name,
         OperationParams.SourceType -> opRead.sourceType
       ))
 
-  private def convertOtherOperation(opOther: op.Operation): DataOperation = {
+  private def findOutputSchema(dataLineage: DataLineage, operation: op.Operation): Map[String, Seq[Map[String, String]]] = {
+    val metaDataset: MetaDataset = dataLineage.datasets.find((dts: MetaDataset) => dts.id == operation.mainProps.output)
+      .getOrElse(throw new IllegalArgumentException(
+        s"Operation output id ${operation.mainProps.output} not found in datasets of dataLineage ${dataLineage.id}"))
+    val attributes = metaDataset.schema.attrs.map(attrId => {
```
wajda: what about algorithmic complexity of this piece?
I'd suggest creating a map of `AttrId -> Attribute` first and then apply it to a `schema.attrs`. E.g.
```suggestion
val attrById = dataLineage.attributes.map(a => a.id -> a).toMap
    val schemaAttributes =
      metaDataset.schema.attrs
        .map(attrById)
        .map(attr => Map(
          "name" -> attr.name,
          "dataTypeId" -> attr.dataTypeId.toString))
```

wajda: Also, using handling options seems unnecessary here. We assume the integrity and correctness of the lineages we process.

PR #299, opened 2019-07-22
spline 296 Spline lineage harvester fails on a particular types of lineages (array literals)

lukasnalezenec:


core/src/test/scala/za/co/absa/spline/core/harvester/ExpressionConverterSpec.scala
```diff
@@ -138,6 +141,15 @@ class ExpressionConverterSpec extends FlatSpec with OneInstancePerTest with Mock
     }
   }
 
+  it should "support array literals" in {
```
wajda: Can you add a few more unit tests to cover more complex types inside a struct?

core/src/main/scala/za/co/absa/spline/core/harvester/componentCreators.scala
```diff
@@ -140,7 +142,11 @@ object ExpressionConverter {
     def render(o: Any): Option[Any] = renderValue(render)(o)
 
     lit.value match {
```
wajda: Avoid `asInstanceOf`. What about doing unapply on a `Literal` instead of a literal value?
```suggestion
lit match {
   case Literal(ad: ArrayData, ArrayType(elementType, _)) => ...
   case Literal(v, _) => ...
}
```

core/src/main/scala/za/co/absa/spline/core/harvester/componentCreators.scala
```diff
@@ -110,7 +111,7 @@ object ExpressionConverter {
     })
   }
 
-  private[this] def renderValue(recursion: Any => Option[Any])(o: Any): Option[Any] = {
+  private[this] def renderValue(recursion: Any => Option[Any], t : Option[org.apache.spark.sql.types.DataType] = None)(o: Any): Option[Any] = {
```
wajda: Prefer adding an appropriate prefix or suffix to an optional valuable. E.g. 'maybeType', 'typeOption' etc.

wajda: As `t` describes the type of `o` they both should logically be in the same parameter group.

core/src/main/scala/za/co/absa/spline/core/harvester/componentCreators.scala
```diff
@@ -140,7 +142,11 @@ object ExpressionConverter {
     def render(o: Any): Option[Any] = renderValue(render)(o)
 
     lit.value match {
-      case ad: ArrayData => ad.toArray(lit.dataType)
+      case ad: ArrayData => {
+        val elementType = lit.dataType.asInstanceOf[ArrayType].elementType
+        val items: Seq[Any] = ad.toArray(elementType).toSeq
```
wajda: Why calling `toSeq` on `Array`?

core/src/main/scala/za/co/absa/spline/core/harvester/componentCreators.scala
```diff
@@ -121,6 +122,7 @@ object ExpressionConverter {
       case map: Map[_, _] => asOption[Map[String, _]](for ((k, v) <- map; r <- recursion(v)) yield k.toString -> r)
       case seq: Traversable[_] => asOption(seq.flatMap(item => recursion(item)).toList)
       case _ if symbol.isModuleClass => Option(symbol.name.toString)
+      case row : GenericInternalRow => Option(row.toSeq(t.get.asInstanceOf[StructType]).flatMap(x => recursion(x)))
```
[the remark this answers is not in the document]
wajda: Prefer using `Option.collect()` to avoid calling `get` with casting.
Also I'm not sure about `flatMap`. Shouldn't it be rather `map' with `orNull`?
```suggestion
case row: GenericInternalRow => t.collect({
    case st: StructType => row.toSeq(st).map(x => recursion(x).orNull)
})
```
Please add more unit tests to cover internal structures and structures with null-fields.

PR #311, opened 2019-08-08
Feature/266 harvester

wajda:


spark/agent/src/main/scala/za/co/absa/spline/harvester/SparkLineageInitializer.scala
```diff
@@ -91,30 +91,21 @@ object SparkLineageInitializer extends Logging {
     def createEventHandler(): Option[QueryExecutionEventHandler] = {
       val configurer = new DefaultSplineConfigurer(defaultSplineConfiguration, sparkSession)
       if (configurer.splineMode != DISABLED) {
```
leborgneromain: Suggestion : (configurer.splineMode != DISABLED) ? createEventHandler(configurer) : None

spark/spark-adapter-api/src/main/scala/za/co/absa/spline/spark/adapter/URIPrefixes.scala
```diff
@@ -14,10 +14,10 @@
...
  */
 
-let _ = require('lodash')
+package za.co.absa.spline.spark.adapter
 
-let getDetails = function (data, nodeId) {
-    return _.find(data, function (details) { return details.mainProps.id === nodeId })
+object URIPrefixes {
+  //prefix used in identifiers for saveAsTable writes
+  val managedTablePrefix = "table://"
```
leborgneromain: Is this one used somewhere ?

spark/agent/src/main/scala/za/co/absa/spline/harvester/operationBuilders.scala
```diff
@@ -110,164 +93,79 @@ class BatchReadNodeBuilder
...
+    schema = Some(outputSchema),
+    params = Map(
+      OperationParams.Name -> operation.nodeName,
+      OperationParams.DestinationType -> operation.format)
   )
 
-  override def ignoreLineageWrite:Boolean = {
-    false
+  override def wasIgnored: Boolean = {
+    writeMetrics.get("numFiles").exists(0.==)
```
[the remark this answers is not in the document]
wajda: It's just a reference to `==` method on `0`, used as a predicate for `exists`. 
It's just another way to write this: 
```scala
writeMetrics.get("numFiles").exists(n => n == 0)
```
It returns `true` if and only if `writeMetrics` contains key `numFiles` with the value equal to `0`
If there is no such key in the metrics then we don't know how many files were touched by the write, and hence we cannot conclude that the write was ignored or not.

spark/spark-adapter-api/src/main/scala/za/co/absa/spline/spark/adapter/URIPrefixes.scala
wajda: yes, it's used in adapters, `SaveAsTableCommandParserImpl` class

spark/agent/src/main/scala/za/co/absa/spline/harvester/SparkLineageInitializer.scala
wajda: I would as well like to do that, but I don't think it's a correct Scala syntax :)

Issue #277, opened 2019-06-18
Swagger doc generation for Producter API


wajda:
Producer API should be sufficiently documented.
Ideally we should Swagger UI on a gateway. If not possible, Swagger Doc generator should be able to generate Swagger def and a static HTML. Then we could host it somewhere.

wajda:
Use model specific types in the producer controllers. Fix Jackson SerDe for Producer

PR #324, opened 2019-08-23
#277 Add api documentation

leborgneromain:


rest-gateway/pom.xml
```diff
@@ -69,6 +69,99 @@
...
         <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>exec-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>gen-swagger-consumer-json</id>
+                        <phase>process-classes</phase>
+                        <goals>
+                        <goal>java</goal>
```
wajda: nit: indent

rest-gateway/pom.xml
```diff
@@ -69,6 +69,99 @@
...
+                        <configuration>
+                            <includePluginDependencies>true</includePluginDependencies>
+                            <executableDependency>
+                                <groupId>za.co.absa.spline</groupId>
+                                <artifactId>rest-api-doc-generator</artifactId>
+                            </executableDependency>
+                            <mainClass>za.co.absa.spline.swagger.SwaggerDocGenCLI</mainClass>
+                            <arguments>
+                                <argument>-o</argument>
+                                <argument>${project.basedir}/src/main/resources/static/consumerSwagger.json</argument>
```
wajda: What about deduplicating `${project.basedir}/src/main/resources/static` part of the path with a custom maven property, say `${swagger.output.path}`?

rest-gateway/src/main/scala/za/co/absa/spline/gateway/rest/DocumentationConfig.scala
```diff
@@ -0,0 +1,36 @@
...
+@EnableWebMvc
+@Configuration
+class DocumentationConfig extends WebMvcConfigurer {
+
+
+  override def addResourceHandlers(registry: ResourceHandlerRegistry) = {
+    registry
+      .addResourceHandler("/**")
+      .addResourceLocations(
+        "classpath:/META-INF/resources/",
```
wajda: Generally classpath resources suppose to contain files needed for the application to work correctly (e.g. config files, mappings, definitions etc), or another static resources if the final build artifact is JAR-file.
In WAR-files however, the content that is accessible from the web should normally be place to the web resources dedicated place (i.e. `src/webapp/*`). Among other reasons there is a security reason - you do not need to serve the content of a classpath or META-INF resources, that is considered sensitive.

.gitignore
```diff
@@ -100,4 +100,10 @@ spark-warehouse/
 metastore_db/
 
 #MacOS
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+# Swagger API documentation generated on build
+rest-gateway/src/main/resources/static/producerSwagger.json
```
wajda: Suggestion: instead of putting the generated output into the `src/` and excluding it explicitly here, you can write it into a separate directory under `target/` (that is already excluded from VCS) and attach it to the project source dir list, or web resources list. This is a common practice in handling generated stuff in Maven.

leborgneromain: Generating direcly into src/main/resources/* let me serve the static html files easilly with spring web according to the reference : https://spring.io/blog/2013/12/19/serving-static-web-content-with-spring-boot

For info, I've tried by curiosity to generate these files outside of ./src/resources/* but instead in ./target/ and it stoped working. 
In the DocumentationConfig.addResourceHandlers, I of course added `registry.addResourceLocations("./target/")`

rest-gateway/pom.xml
leborgneromain: Done

rest-gateway/pom.xml
leborgneromain: Done

rest-gateway/src/main/scala/za/co/absa/spline/gateway/rest/DocumentationConfig.scala
leborgneromain: ok, removed the resourceLocation of META-INF (this is not needed anyway).

Issue #325, opened 2019-08-27
Some integration tests fail on Windows


benedeki:
Some tests compose Spark SQL command disregarding the fact that Windows path delimited is `\` and don't escape those. This makes the tests fail.

PR #326, opened 2019-08-27
#325: Using toUri to circumvent issues with backslash as

benedeki:


integration-tests/src/test/scala/za/co/absa/spline/BasicIntegrationTests.scala
```diff
@@ -91,13 +92,15 @@ class BasicIntegrationTests extends FlatSpec
...
+
+
   "saveAsTable" should "use table path as identifier when writing to external table" in
     withNewSparkSession(spark =>
       withLineageTracking(spark) {
         lineageCaptor => {
           val path = TempDirectory("sparkunit", "table", pathOnly = true).deleteOnExit().path
 
-          spark.sql(s"create table e_table(num int) using parquet location '$path'")
+          spark.sql(s"create table e_table(num int) using parquet location '${path.toUri}'")
```
Zejnilovic: Is there a reason to call `toUri` here and not when creating the path val like on line 74?

benedeki: Yes, `path` is used further in the code (line 111)

Zejnilovic: Ah I see the `toFile`

PR #343, opened 2019-09-05
Bugfix/spline 340 minor fixes

wajda:


parent/pom.xml
```diff
@@ -263,7 +263,7 @@
                         <artifactId>apache-rat-plugin</artifactId>
                         <executions>
                             <execution>
-                                <phase>verify</phase>
+                                <phase>validate</phase>
```
leborgneromain: Thanks for this.

Issue #345, opened 2019-09-06
File separator issues prevent build on Windows


benedeki:
While in most cases Windows nowadays accept `/` as the file separator, sometimes `\` is still needed.
On few plays of the project files the hard-coded `/` prevents build on Windows system. In those cases using `${file.separator}` fixes the issue.

wajda:
Can you please give us more details about the issue? What hardcoded slashed and were prevent build?

benedeki:
Most of build tools actually works OK. But some plugins require OS specific file separator.
E.g.
`rest-gateway/pom.xml` line 144
```
<exec executable="cmd" osfamily="windows" failonerror="true">
  <arg line="/c node_modules${file.separator}.bin${file.separator}redoc-cli.cmd bundle -o ${swagger.output.path}/consumer.html ${swagger.output.path}/consumerSwagger.json"/>
</exec>
```

leborgneromain:
@benedeki I found this bug already and it will be fixed in the next PR.

wajda:
Since Romain got it fixed in his PR already (4ff6db8a7ed82fc5fab1d17eb823378b79e429e5, specifically [here](https://github.com/AbsaOSS/spline/pull/347/files#diff-387b4088e123ef74679c967e667c6c07R144) ) I'd suggest to close it.

PR #346, opened 2019-09-06
#345: File separator issues prevent build on Windows

benedeki:
* Using ${file.separator} instead of / on the critical places

### DIFF
--- a/client-ui/assembly.xml
+++ b/client-ui/assembly.xml
@@ -25,7 +25,7 @@
   <fileSets>
     <fileSet>
       <directory>${project.basedir}/dist/ui2</directory>
-      <outputDirectory>/</outputDirectory>
+      <outputDirectory>./</outputDirectory>
     </fileSet>
   </fileSets>
 </assembly>
--- a/client-ui/package-lock.json
+++ b/client-ui/package-lock.json
@@ -1439,7 +1439,7 @@
     },
     "@types/node": {
       "version": "8.9.5",
-      "resolved": "http://registry.npmjs.org/@types/node/-/node-8.9.5.tgz",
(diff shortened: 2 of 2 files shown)


leborgneromain:
LGTM

wajda:
I might be wrong, but I thing using `${file.separator}` is an overkill. It makes plugin config harder to read, and is not equally necessary in different places.
For instance, in the Ant plugin commands are split by OS, so you can just use Windows backslash there:
```xml
<exec executable="cmd" osfamily="windows" failonerror="true">
    <arg line="/c node_modules\.bin\redoc-cli.cmd bundle -o ${swagger.output.path}\consumer.html ${swagger.output.path}\consumerSwagger.json"/>
    ...
</exec>
```  
In the other place - the Tomcat plugin - `<path>/</path>` elements denotes the Web-context name (root context n this case) - it;s a URI part, not a file system path. So, using backslash could event hurt here.

benedeki:
>
>
> I think using `${file.separator}` is an overkill. It makes plugin config harder to read, and is not equally necessary in different places.
> For instance, in the Ant plugin commands are split by OS, so you can just use Windows backslash there:

I used it on the minimum required places needed. Sure the backslash can be used, but seems more brittle depending on the designation 
 
> In the other place - the Tomcat plugin - `<path>/</path>` elements denotes the Web-context name (root context n this case) - it;s a URI part, not a file system path. So, using backslash could event hurt here.

In this case the build actually eventually succeeds  But there's a big fat red ERROR in the build log. This removes it. Reportedly using `./` also works if you don't like `${file.separator}`.
Let's discuss Monday.

benedeki:
The change is very different from the original PR. But it does address an error from the build log.
Second is a change in the `package-lock.json` in hope that after this commit it will remain stable.

PR #359, opened 2019-09-20
Bugfix/spline 306 fix source uris

wajda:
This PR addresses issues with identifying datasources when reading from and writing to using different Spark methods.
In particular it addresses support for managed Spark/Hive tables and JDBC read/write.
It also obsoletes and removes Spark and Json4s adapter modules. Support for different versions of Spark and Json4s libraries are done via reflection, all in one place.


spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/WriteCommandExtractor.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  private def asTableWriteCommand(table: CatalogTable, mode: SaveMode, query: LogicalPlan) = {
+    WriteCommand(CatalogTableUtils.getSourceIdentifier(table)(pathQualifier, session), mode, query)
+  }
+}
+
+object WriteCommandExtractor {
+
+  private object `_: InsertIntoHiveTable` extends SafeTypeMatchingExtractor(classOf[InsertIntoHiveTable])
```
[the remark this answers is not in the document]
wajda: The object name is specifically crafted to resemble a conventional Scala type pattern matching syntax when is used in the `match` block as a syntactic sugar. This extractor object semantic is almost identical to the Scala's native analog with the only exception - it doesn't fail with a fatal exception if the checking type isn't in the classpath.

commons/src/test/scala/za/co/absa/spline/common/ReflectionUtilsSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  class MyClass extends MyTrait
+
+  object MyObject extends MyTrait
+
+  case class Foo(x: String, y: Int)
+
+  object Foo {
+    //noinspection ScalaUnusedSymbol
+    private[this] val privateVal = 42
```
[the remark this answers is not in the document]
wajda: used in a test

commons/src/main/scala/za/co/absa/spline/common/extractors/SafeTypeMatchingExtractor.scala
```diff
@@ -14,15 +14,18 @@
...
 
-package za.co.absa.spline.json4s.adapter
+package za.co.absa.spline.common.extractors
 
-import java.text.SimpleDateFormat
+class SafeTypeMatchingExtractor[T](lazyClass: => Class[T]) {
+  private val classIfAvailable: Option[Class[_]] =
+    try Some(lazyClass)
+    catch {
+      case _: NoClassDefFoundError | _: ClassNotFoundException => None
```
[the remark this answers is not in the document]
wajda: done in purpose

commons/src/test/scala/za/co/absa/spline/common/extractors/AccessorMethodValueExtractorSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+  }
+}
+
+object AccessorMethodValueExtractorSpec {
+
+  trait Wrong
+
+  case class TestObj(x: Int, map: Map[_, _])
+
+  private val testObj = TestObj(42, Map.empty)
```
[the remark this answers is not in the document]
wajda: used in tests

commons/src/main/scala/za/co/absa/spline/common/extractors/SafeTypeMatchingExtractor.scala
```diff
@@ -14,15 +14,18 @@
  * limitations under the License.
  */
 
-package za.co.absa.spline.json4s.adapter
+package za.co.absa.spline.common.extractors
 
-import java.text.SimpleDateFormat
+class SafeTypeMatchingExtractor[T](lazyClass: => Class[T]) {
+  private val classIfAvailable: Option[Class[_]] =
+    try Some(lazyClass)
```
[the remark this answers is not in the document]
wajda: must be `try {}` to catch fatal exceptions

commons/src/test/scala/za/co/absa/spline/common/ReflectionUtilsSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+  sealed trait MyTrait
+
+  class MyClass extends MyTrait
+
+  object MyObject extends MyTrait
+
+  case class Foo(x: String, y: Int)
+
+  object Foo {
+    //noinspection ScalaUnusedSymbol
```
leborgneromain: to remove I guess

parent/pom.xml
```diff
@@ -346,6 +343,12 @@
             <artifactId>scalatest_${scala.compat.version}</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
leborgneromain: where this one is use in the code ? I am missing something I think

spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/CatalogTableUtils.scala
```diff
@@ -0,0 +1,38 @@
...
+object CatalogTableUtils {
+  def toSourceIdentifier(table: CatalogTable)
+    (pathQualifier: PathQualifier, session: SparkSession): SourceIdentifier = {
+    val uri = table.storage.locationUri
+      .map(_.toString)
+      .getOrElse({
+        val catalog = session.catalog
+        val TableIdentifier(tableName, maybeTableDatabase) = table.identifier
+        val databaseName = maybeTableDatabase getOrElse catalog.currentDatabase
+        val databaseLocation = catalog.getDatabase(databaseName).locationUri.stripSuffix("/")
```
leborgneromain: suggestion : can't you remove stripSuffix("/") and just concatenate the databaseLocation directly ?
Or maybe it's more clear like this I don't know

spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/WriteCommand.scala
```diff
@@ -14,10 +14,13 @@
...
 
+import org.apache.spark.sql.SaveMode
 import org.apache.spark.sql.catalyst.plans.logical.{Command, LogicalPlan}
 
-abstract class AbstractWriteCommand extends Command {
-  val query: LogicalPlan
-}
+case class WriteCommand(
+  sourceIdentifier: SourceIdentifier,
+  mode: SaveMode,
```
leborgneromain: This is clearer but we have a boolean on the persistence model atm, it needs to be modified and few other services as well I think

spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/WriteNodeBuilder.scala
```diff
@@ -0,0 +1,45 @@
...
+  extends OperationNodeBuilder {
+
+  override protected type R = WriteOperation
+  override val operation: LogicalPlan = command.query
+
+  override def build(): WriteOperation = {
+    val Seq(uri) = command.sourceIdentifier.uris
+    WriteOperation(
+      outputSource = uri,
+      append = command.mode == SaveMode.Append,
```
leborgneromain: Maybe create a new issue for this because it has few impacts in the persistence and services if we want to map it correctly. Maybe it's solved in your next PRs ?

spark/agent/src/main/scala/za/co/absa/spline/harvester/json/ShortTypeHintForSpline03ModelSupport.scala
```diff
@@ -0,0 +1,48 @@
...
+import scala.reflect.runtime.universe._
+
+trait ShortTypeHintForSpline03ModelSupport extends FormatsBuilder {
+  override protected def formats: Formats = createFormats(Map(
+    "typeHintFieldName" -> "_typeHint",
+    "typeHints" -> ShortTypeHints(
+      subClassesOf[model.dt.DataType] ++
+        subClassesOf[model.expr.Expression]
+    ),
+    "dateFormatter" -> DefaultFormats.losslessDate.get))
```
leborgneromain: The bot is right here, no ?

commons/src/main/scala/za/co/absa/spline/common/ReflectionUtils.scala
```diff
@@ -17,7 +17,9 @@
 package za.co.absa.spline.common
 
 import scala.collection.concurrent.TrieMap
+import scala.reflect.runtime.universe._
 import scala.reflect.runtime.{universe => ru}
```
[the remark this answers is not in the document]
wajda: The second is a package alias. But you're right, it's seems unnecessary in the context.

commons/src/test/scala/za/co/absa/spline/common/ReflectionUtilsSpec.scala
wajda: Well, actually it's an IDEA specific comment to disable that type of inspection on the statement. It has no effect on Scala compiler, but IDEA is cleaver enough to complain about unused `private[this] val privateVal = 42`, which is in fact accessed via reflection in a test method.

parent/pom.xml
wajda: Do you mean jsr305? It's an optional library that contain annotations that one can use in the code to assist automated bug detection tools. We don't really use it (hence scope provided), but it is used in many other libraries that we use, and without in being available at compile time Scala compiler generates warnings. So, adding it as a globally provided library for the project is merely a measure to minimize the amount of false positive warning during the build process.

spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/CatalogTableUtils.scala
wajda: I'm stripping an optional trailing slash before concatenation to prevent missing slashed or double slash occurrences after concatenation.

spark/agent/src/main/scala/za/co/absa/spline/harvester/json/ShortTypeHintForSpline03ModelSupport.scala
wajda: No, the bot is being stupid here :) The type is `ThreadLocal[SimpleDateFormat]`

spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/WriteNodeBuilder.scala
wajda: as discussed offline there is no impact on the model.

PR #372, opened 2019-10-07
Spline #371 Use caseSensitive map instead of spark CaseInsensitiveMap…

leborgneromain:
… for readOperation params Map

### DIFF
--- a/spark/agent/src/test/scala/za/co/absa/spline/harvester/builder/BuilderSpec.scala
+++ b/spark/agent/src/test/scala/za/co/absa/spline/harvester/builder/BuilderSpec.scala
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2019 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


wajda:
The same fix has also to be done for other builders, namely `WriteNodeBuilder` and `GenericNodeBuilder`.

spark/agent/src/test/scala/za/co/absa/spline/harvester/builder/BuilderSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.scalatest.mockito.MockitoSugar
+import org.scalatest.{FlatSpec, Matchers}
+import za.co.absa.spline.harvester.ComponentCreatorFactory
+import za.co.absa.spline.harvester.builder.read.{ReadCommand, ReadNodeBuilder}
+
+class BuilderSpec extends FlatSpec with Matchers with MockitoSugar {
+
+  implicit private val componentCreatorFactory: ComponentCreatorFactory = new ComponentCreatorFactory
+
+  private val logicalPlanMock = mock[LogicalPlan]
```
wajda: minor: I'd rather call the variable _stub_ or _dummy_ instead of _mock_, as it's corresponds its purpose more.

Mocks are usually more focused on behavior and observing the actions, kinda are more involved into the unit test logic. While stubs and dummies are more for test isolation purposes, e.g. dumb entities preset with values/answers to allow the testing code not to break because of missing side dependencies that are _not_ being directly tested in the given test.
```suggestion
  private val logicalPlanStub = mock[LogicalPlan]
```

PR #391, opened 2019-10-16
add test that uses createOrReplaceTempView and stores data in hive table

cerveada:


integration-tests/src/test/scala/za/co/absa/spline/harvester/LineageHarvesterSpec.scala
```diff
@@ -218,6 +226,56 @@ class LineageHarvesterSpec extends FlatSpec with Matchers with SparkFixture with
...
+          firstOperation.childIds shouldEqual Seq(2)
+          firstOperation.params("name") shouldEqual "Project"
+
+          val secondOperation = otherOperations(1)
+          secondOperation.id shouldEqual 2
+          secondOperation.childIds shouldEqual Seq(3)
+          if (ver"$SPARK_VERSION" < ver"2.4")
+            secondOperation.params("name") shouldEqual "SubqueryAlias"
+          else
+            secondOperation.params("name") shouldEqual Some("`tempview`")
```
wajda: Common unit tests design practices suggest to avoid `if-else` statements as well as other types of branching in general. The reason is that it puts cognitive load onto the reader, while unit tests are supposed to be easily readable.
I'd suggest using disjunctive assertion instead of branching.
```suggestion
            secondOperation.params("name") should (equal("SubqueryAlias") or equal("tempview"))
```

wajda:
please squash and merge

wajda:
and please make sure you add the related issue number into every commit message, so that every commit is linked with the issue.

image

Issue #387, opened 2019-10-11
Log a warning when read or write command is not implemented.


cerveada:
Currently, when ```WriteCommandExtractor``` or ```ReadCommandExtractor``` are not able to handle certain command they will silently return a None and the lineage will be lost.

Instead, error should be thrown or at least it should be logged.

cerveada:
* The commands that have no influence on lineage can remain ignored.
* The commands that influence a lineage will be all processed by spline in the future, but for now they should be at least logged, so the user know this is part of not yet implemented functionality. Test (for a future use) should be created to produce the command and validate that lineage for it was caught.

PR #408, opened 2019-10-23
spline #387 log a warning for unimplemented spark commands

cerveada:
Also adds test to produce the unimplemented commands.


spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteCommandExtractor.scala
```diff
@@ -98,6 +107,18 @@ class WriteCommandExtractor(pathQualifier: PathQualifier, session: SparkSession)
     val sourceIdentifier = SourceIdentifier.forTable(table)(pathQualifier, session)
     WriteCommand(name, sourceIdentifier, mode, query, Map("table" -> table))
   }
+
+  private val toBeImplemented = Seq(classOf[AlterTableAddColumnsCommand], classOf[AlterTableChangeColumnCommand],
+    classOf[AlterTableRenameCommand], classOf[AlterTableSetLocationCommand], classOf[CreateDataSourceTableCommand],
+    classOf[CreateDatabaseCommand], classOf[CreateTableLikeCommand], classOf[DropDatabaseCommand],
+    classOf[LoadDataCommand], classOf[TruncateTableCommand])
+
+  private def logUnimplementedComamnd(c: LogicalPlan): Unit = {
```
wajda: What if we vary the behavior of this method depending on the `SplineMode` property?
So that in the _BestEffort_ mode we'd log a warning, but in the _Required_ mode we throw an exception instead.

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
```diff
@@ -0,0 +1,285 @@
...
+          val (plan, _) = lineageCaptor.lineageOf {
+            spark.sql(s"TRUNCATE TABLE $tableName") // TruncateTableCommand
+          }
+        }
+      }
+    }
+
+  "Lineage for alter table add columns" should "be caught" in
+    withCustomSparkSession(_
+    .enableHiveSupport()
```
wajda: nit: indent

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
```diff
@@ -0,0 +1,285 @@
...
+
+  "Lineage for load data" should "be caught" in
+    withCustomSparkSession(_
+      .enableHiveSupport()
+      .config("hive.exec.dynamic.partition.mode", "nonstrict")) { spark =>
+
+      withHiveDatabase(spark)(databaseName,
+        (tableName, "(x String, ymd int)", Seq(("Tata", 20190401), ("Tere", 20190403)))) {
+
+        val filePath = tempDirPath + "/loadData.txt"
```
wajda: we prefer string interpolation to string concatenation

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
```diff
@@ -0,0 +1,285 @@
...
+          StructField("ID", IntegerType, nullable = false) :: StructField("NAME", StringType, nullable = false) :: Nil)
+
+        val rdd = spark.sparkContext.parallelize(Row(1014, "Warsaw") :: Row(1002, "Corte") :: Nil)
+        spark.sqlContext.createDataFrame(rdd, schema)
+      }
+
+      testData.write.jdbc(jdbcConnectionString, "atable", new Properties)
+
+      spark.sql(
+        s"CREATE TABLE jdbcTable USING org.apache.spark.sql.jdbc OPTIONS (" +
```
wajda: Please use a triple-quotes for multiline strings
```scala
"""
  XXXX
  YYY
  ZZZZ
"""
```
or 
```scala
"""
  |XXXX
  |YY
  |ZZZZ
  |""".stripMargin
```

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
```diff
@@ -0,0 +1,285 @@
...
+  with Matchers
+  with SparkFixture
+  with SplineFixture
+  with SparkDatabaseFixture
+  with JDBCFixture {
+
+  val databaseName = "testDb"
+  val tableName = "testTable"
+
+  private val tempDirPath = TempDirectory(prefix = "test").deleteOnExit().path.toFile.getAbsolutePath
```
wajda: `tempDirPath` is only used once and the usage is quite far away from the definition. Is there any reason why it shouldn't be put closer to the usage, or even be inlined?

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
```diff
@@ -0,0 +1,285 @@
...
+
+        withLineageTracking(spark) { lineageCaptor =>
+          val (plan, _) = lineageCaptor.lineageOf {
+            spark.sql(s"LOAD DATA LOCAL INPATH '$filePath' INTO TABLE $tableName") // LoadDataCommand
+          }
+        }
+      }
+    }
+
+  private def printToFile(filePath: String)(op: java.io.PrintWriter => Unit) {
```
wajda: Please take a look at `org.apache.commons.io.FileUtils` class. It should contain a method that you can reuse.

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
```diff
@@ -0,0 +1,285 @@
...
+      .enableHiveSupport()
+      .config("hive.exec.dynamic.partition.mode", "nonstrict")) { spark =>
+        try
+          withLineageTracking(spark) { lineageCaptor =>
+            val (plan, _) = lineageCaptor.lineageOf {
+              spark.sql(s"CREATE DATABASE $databaseName") // CreateDatabaseCommand
+            }
+          }
+        finally
+          spark.sql(s"DROP DATABASE $databaseName CASCADE")
```
wajda: Usually a test method shouldn't be responsible for cleanup, to keep the tests more concise and focused. The cleanup should rather be done in one the fixtures or `before/after` methods.

In this particular case you can use `withRestartingSparkContext` loan-method that would restart Spark and do all the cleanup for you.

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
```diff
@@ -0,0 +1,285 @@
...
+
+        printToFile(filePath) { p =>
+          val separator = "\u0001"
+          p.println("FooBar" + separator + "42")
+          p.println("BleBla" + separator + "66")
+        }
+
+        withLineageTracking(spark) { lineageCaptor =>
+          val (plan, _) = lineageCaptor.lineageOf {
+            spark.sql(s"LOAD DATA LOCAL INPATH '$filePath' INTO TABLE $tableName") // LoadDataCommand
```
wajda: We should be careful when inserting `Path`'s string representation verbatim into SQL. It might lead to malformed SQL statement for certain paths. E.g. on Windows it might use backslashes that must be escaped in SQL.
One way of solving the issue is to rely on `URI.toString` instead of `Path.toString`. See `BasicIntegrationTests.scala:129`

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
```diff
@@ -0,0 +1,285 @@
...
+
+  "Lineage for alter table set location" should "be caught" in
+    withCustomSparkSession(_
+      .enableHiveSupport()
+      .config("hive.exec.dynamic.partition.mode", "nonstrict")) { spark =>
+
+      withHiveDatabase(spark)(databaseName,
+        (tableName, "(x String, ymd int)", Seq(("Tata", 20190401), ("Tere", 20190403)))) {
+
+        val newPath = getNewTableLocation(spark)
```
wajda: Why don't you simply use `TempDirectory` to create a new path for a table?

spark/agent/README.md
```diff
@@ -0,0 +1,100 @@
+# Spark Agent / Harvester
+
+This module is responsible for listening to spark command events and converting them to spline lineage.
+
+## Spark commands
+Some events provided by Spark are not yet implemented. Some of them will be implemented in future 
+and some of them bear no lineage information and should be ignored.
+
+All commands inherits from ```org.apache.spark.sql.catalyst.plans.logical.Command```.
```
Zejnilovic:
```suggestion
All commands inherit from ```org.apache.spark.sql.catalyst.plans.logical.Command```.
```

spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteCommandExtractor.scala
```diff
@@ -19,73 +19,84 @@ package za.co.absa.spline.harvester.builder.write
...
+class WriteCommandExtractor(pathQualifier: PathQualifier, session: SparkSession, splineMode: SplineMode) {
+
+  private val logger = LoggerFactory.getLogger(getClass.getSimpleName)
+
+  def asWriteCommand(operation: LogicalPlan): Option[WriteCommand] = {
+    val maybeWriteCommand = condOpt(operation) {
+      case cmd: SaveIntoDataSourceCommand =>
+        val maybeSourceType = DataSourceTypeExtractor.unapply(cmd)
+        maybeSourceType match {
+          case Some(sourceType) if sourceType == "jdbc" || sourceType.isInstanceOf[JdbcRelationProvider] =>
```
Zejnilovic: These are quite big, do you think extracting some of the cases would make the code more readable?

spark/agent/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+You can see how to produce unimplemented commands in ```za.co.absa.spline.harvester.SparkUnimplementedCommandsSpec```.
+### Implemented
+```
+CreateDataSourceTableAsSelectCommand (org.apache.spark.sql.execution.command)
```
Zejnilovic: why not use plain `-` (\<ul>)?

cerveada: I usually use ``` to mark that the word is a piece of code or specific Scala class. Here it's all classes, so following the same logic I just put them all in one block.

spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteCommandExtractor.scala
wajda: I'd say it's alright. Considering the fact that the match is far from being complete and that there is (still) a large unknown area for us in terms of finding a right implementation of those cases (existing ones and the ones that will be added in next versions), I'd prefer to keep them close together to have better view on all at once, and to facilitate the ongoing refactoring during further development. We'll probably decompose this class later when we have better understanding of how those pieces should interact, what can/should be reused and so forth. For now it's in the state of ongoing development.

spark/agent/README.md
Zejnilovic: Yes, I agree 100% on marking code with ` or ``` depending on the situation. But I would expect this.

- `CreateHiveTableAsSelectCommand` (org.apache.spark.sql.hive.execution)
- `CreateTableCommand` (org.apache.spark.sql.execution.command)


In no way do I want to force you into it. Just give it a thought

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
```diff
@@ -44,20 +47,18 @@ class SparkUnimplementedCommandsSpec extends FlatSpec
...
     withCustomSparkSession(_
       .enableHiveSupport()
       .config("hive.exec.dynamic.partition.mode", "nonstrict")) { spark =>
-        try
-          withLineageTracking(spark) { lineageCaptor =>
-            val (plan, _) = lineageCaptor.lineageOf {
-              spark.sql(s"CREATE DATABASE $databaseName") // CreateDatabaseCommand
-            }
+
+      withRestartingSparkContext {
```
wajda: `withRestartingSparkContext` method is designed to be used _before_ `withCustomSparkSession` because it kills the session and effectively undoes all the changes to the session in all previous methods.
So it this test Hive is not actually enabled. Do you actually need it considering that the test passes without Hive?

spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteCommandExtractor.scala
```diff
@@ -112,17 +110,29 @@ class WriteCommandExtractor(pathQualifier: PathQualifier, session: SparkSession)
...
+    classOf[LoadDataCommand],
+    classOf[TruncateTableCommand]
+  )
+
+  private def alertWhenUnimplementedCommand(c: LogicalPlan): Unit = {
+    if (commandsToBeImplemented.contains(c.getClass)) {
+      val msg = s"Spark command was intercepted, but is not yet implemented! Command:'${c.getClass}'"
+
+      splineMode match {
+        case SplineMode.REQUIRED => throw new UnsupportedOperationException(msg)
```
wajda: Can you please define a new checked exception class for this purpose and use that one instead?
Also please declare it in the method signature.
`UnsupportedOperationException` is too generic and might be thrown from underlying code in Spark or potentially other libraries. Also as it's not declared in the class contract the caller can easily miss it.

spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteCommandExtractor.scala
```diff
@@ -35,57 +36,65 @@ import scala.language.reflectiveCalls
...
-      val sourceId = SourceIdentifier(None, pathQualifier.qualify(uri))
-      WriteCommand(dtc.nodeName, sourceId, Overwrite, dtc)
-
-    case ctc: CreateTableCommand =>
-      val sourceId = SourceIdentifier.forTable(ctc.table)(pathQualifier, session)
-      WriteCommand(ctc.nodeName, sourceId, Overwrite, ctc)
+  private val logger = LoggerFactory.getLogger(getClass.getSimpleName)
+
+  def asWriteCommand(operation: LogicalPlan): Option[WriteCommand] = {
+    val writeCommand = condOpt(operation) {
```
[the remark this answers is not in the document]
cerveada: ok

integration-tests/src/test/scala/za/co/absa/spline/harvester/SparkUnimplementedCommandsSpec.scala
cerveada: I used database fixture instead, that should be quicker and take care of the cleaning.
The Hive is actually not needed in this case, so I removed it, but in most of other tests it's necessary and also useful for creating the tests.

Issue #405, opened 2019-10-22
UI doesn't show any error when Consumer is inaccessible


wajda:
When Consumer is not running XHR response status code is 0, and it apparently either isn't caught by our error handler, or is interpreted differently. No error is shown on UI and everything looks like there's just no data.  

![image](https://user-images.githubusercontent.com/795479/67307998-b7407680-f4f9-11e9-85d6-a35c5e8259e6.png)

Zejnilovic:
LGTM

PR #423, opened 2019-10-29
Spline #405 Add exception handler for execution events http calls + r…

leborgneromain:
…efactor of handleError function


client-ui/src/app/effects/execution-events.effects.ts
```diff
@@ -48,7 +48,11 @@ export class ExecutionEventsEffects {
     public getPageableExecutionEvents$: Observable<Action> = this.actions$.pipe(
         ofType(ExecutionEventsAction.ExecutionEventsActionTypes.EXECUTION_EVENTS_GET),
         switchMap((action: any) => this.executionEventControllerService.executionEventUsingGET(action.payload)),
-        map((res: PageableExecutionEvent) => new ExecutionEventsAction.GetSuccess(res))
+        map((res: PageableExecutionEvent) => new ExecutionEventsAction.GetSuccess(res)),
+        catchError(err => {
+            this.store.dispatch(new ErrorActions.ServiceErrorGet(handleError(err)))
+            return empty()
+        })
```
cerveada: This error handling fragment is here several times, would it make sense to refactor it to function?

leborgneromain: Good point, it should be possible by creating a custom rxjs operator, I will try to add it but I need to read a bit about it first.

Issue #409, opened 2019-10-23
Incorrect total count on the dashboard


wajda:
The total count of jobs on the dashboard is limited to 100.
If I'm looking for a job that doesn't fall into this bucket then the job isn't found.

Example:

![image.png](https://images.zenhubusercontent.com/5c04fd55820d8748e06666cd/8a29f8c2-320d-4b33-8fc4-97ac02db52fa)

Zejnilovic:
In case I ran new jobs that made lineage and I press the go to the beginning of a list, the count will drop by the number of new destinations written. So old ones are removed new ones are not added.

wajda:
@Zejnilovic 
> In case I ran new jobs that made lineage and I press the go to the beginning of a list, the count will drop by the number of new destinations written. So old ones are removed new ones are not added.

Can you please retest it on Develop?

wajda:
> In case I ran new jobs that made lineage and I press the go to the beginning of a list, the count will drop by the number of new destinations written. So old ones are removed new ones are not added.

@Zejnilovic Could you please retest this and close the issue if it works?

Zejnilovic:
> > In case I ran new jobs that made lineage and I press the go to the beginning of a list, the count will drop by the number of new destinations written. So old ones are removed new ones are not added.
> 
> @Zejnilovic Could you please retest this and close the issue if it works?

I would say this is ok since there is no page limit and no automatic refresh of jobs.

Issue #351, opened 2019-09-12
Pre-release refactoring: REST and Web layers


wajda:
Rename (if necessary, for the sake of consistency and descriptiveness) the REST API URLs, improve Swagger docs and other things that are a part of public API

wajda:
Rename `spark-bundle` module to `spline-spark-agent-bundle` or something more descriptive.

PR #439, opened 2019-11-02
spline #409 Incorrect total count on the dashboard

wajda:


consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/ExecutionEventController.scala
```diff
@@ -41,7 +41,7 @@ class ExecutionEventController @Autowired()(val repo: ExecutionEventRepository)
     @ApiParam(value = "Beginning of the timestamp range used for querying. If timestampStart equals 0, the service will return the first 100 execution events in database", example = "0")
     @RequestParam(value = "timestampStart", required = false, defaultValue = "0") timestampStart: Long,
     @ApiParam(value = "End of the timestamp range used for querying", example = "0")
-    @RequestParam(value = "timestampEnd", required = false, defaultValue = "0") timestampEnd: Long,
+    @RequestParam(value = "timestampEnd", required = false, defaultValue = "9223372036854775807") timestampEnd: Long,
```
cerveada: Might be better to define a value for his number or comment it.

wajda: It's a max Long :)
I though about making it a constant, but then I'd have to extract the rest of defaults into constants. Just hesitated to do that and inlined for the sake of consistency. Let's me handle it in the scope of #351

cerveada: ok, In that case you could also do `Long.MaxValue.toString`

wajda: Unfortunately I couldn't. The annotation value must be a constant expression. Calling `.toString` makes it dynamic.

PR #465, opened 2019-11-15
Feature/ui modifications

leborgneromain:


client-ui/src/app/components/time-frame-picker/date-picker/date-picker.component.html
```diff
@@ -14,24 +14,7 @@
   ~ limitations under the License.
   -->
 
-<form class="form-inline">
-  <div class="form-group">
-    <div class="input-group">
-      <input name="dp" ngbDatepicker
-             class="form-control" [class.is-invalid]="!valid"
```
wajda: I see `[class.is-invalid]="!valid"` is missing in your variant. Is it intentional? It was there for visual indication of invalid input.

client-ui/src/app/components/time-frame-picker/date-picker/date-picker.component.html
```diff
@@ -14,24 +14,7 @@
...
-
-<ng-template #footerTemplate>
-  <hr class="my-0">
-  <button class="btn btn-primary btn-sm m-2 float-left" (click)="onTodayClick(); d.close()">Today</button>
-  <button class="btn btn-secondary btn-sm m-2 float-right" (click)="d.close()">Close</button>
-</ng-template>
+<input type="text" class="form-control" placement="top" #dp="bsDatepicker" bsDatepicker [bsValue]="bsModel"
+  [minDate]="bsMinDate" [maxDate]="bsMaxDate"
+  [bsConfig]="{  dateInputFormat: 'DD/MM/YYYY', containerClass: 'theme-dark-blue' }"
+  value="{{ bsModel | date: 'dd/MM/yyyy' }}" (bsValueChange)="onModelChange($event)">
```
wajda: That's probably a matter of taste, but without a proper localization support I'd suggest to stay away from `dd/MM/yyyy` format, as it's easily confused with american `MM/dd/yyyy`. I'd suggest using Swedish date format `YYYY-MM-DD`. It's the least confusing and is easily understandable in majority of countries I think. Hence the best candidate for "universal" date format IMO.

Issue #496, opened 2019-12-02
Add "Operation.extras" to avoid possible property name collision


wajda:
Operation property `name` could mean different thing for certain operations. The same states for the other Agent specific properties that we mix-in into the `Operation.params`
We agreed to add `Operation.extras` key-value set and move those extra properties there.

PR #505, opened 2019-12-03
spline #496 Add "Operation.extra" to avoid possible property name col…

wajda:
…lision


spark/agent/src/main/scala/za/co/absa/spline/harvester/builder/write/WriteNodeBuilder.scala
```diff
@@ -37,10 +37,11 @@ class WriteNodeBuilder(command: WriteCommand)
       append = command.mode == SaveMode.Append,
       id = id,
       childIds = childIds,
-      params = Map(
-        OperationParams.Name -> command.name,
-        OperationParams.DestinationType -> command.sourceIdentifier.format
-      ) ++ command.params
+      params = Map(command.params.toSeq: _*),
```
cerveada: Why create new map here?

wajda: Because Spark returns a case insensitive map that created issues in the past. See #371

PR #516, opened 2019-12-07
bug/spark-515 added missing catch statement to HttpLineageDispatcher.tryStatusOk

calvertj:
simple change, a type should probably be added as well..

### DIFF
--- a/spark/agent/src/test/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcherSpec.scala
+++ b/spark/agent/src/test/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcherSpec.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2019 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


wajda:
Wow, that's a good one! :thumbsup: :smile: 
Could you also fix the tests for that?

cerveada:
I will take care of the test.

spark/agent/src/test/scala/za/co/absa/spline/harvester/dispatcher/HttpLineageDispatcherSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+import scalaj.http._
+import za.co.absa.spline.harvester.exception.SplineNotInitializedException
+
+class HttpLineageDispatcherSpec extends AnyFlatSpec with MockitoSugar {
+
+  behavior of "HttpLineageDispatcher"
+
+  val dummyUrl = "http://dummyUrl"
+
+  val httpMock = mock[BaseHttp]
```
wajda: Minor: Even though it's a unit-test these val's had better to be marked as `private`. At least because it removes those IDE's warnings (otherwise it complains about missing type annotation on public members)

cerveada:
To release this as a hotfix for 0.4. I rebased the commits from here to #519. Once new 0.4 version is released we will back merge the fixes to develop.

Issue #406, opened 2019-10-22
Migrator cli - add validations


Zejnilovic:
- [ ] -r file must exist
- [ ] -l loglevel is one of `ALL`, `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR`, `OFF`
- [ ] -e file should not exist -> This is mainnly because I don't see a feature of removing from file what has already been "recovered"

Zejnilovic:
scopt generally doesn't have validations. Maybe do a scopt-wide update?

wajda:
Since we removed Migrator tool from 0.5 I'll move the changes to 0.4 branch. If we need to do other fixes in Migrator we'll either do it in 0.4.x or even extract it as a separate project.

PR #546, opened 2020-01-07
spline #406 migrator cli input validations

cerveada:


migrator-tool/src/main/scala/za/co/absa/spline/migrator/MigratorCLI.scala
```diff
@@ -57,11 +62,13 @@ class MigratorCLI(migratorTool: MigratorTool) {
         """|A file where a list of failed lineage IDs will be written to.
            |Running migrator with the option '-r' will repeat attempt to migrate lineages from this file."""
           .stripMargin
+        validate (f => if (!f.exists()) success else failure("<failrec> file should not yet exist"))
```
Zejnilovic: I think it should be more like `Failrec file with specified path and name already exists` or just `file already exists`. Say what is wrong, don't say what should (not) be.

migrator-tool/src/main/scala/za/co/absa/spline/migrator/MigratorCLI.scala
```diff
@@ -35,6 +35,11 @@ object MigratorCLI extends App {
 }
 
 class MigratorCLI(migratorTool: MigratorTool) {
+
+  private val logLevels = Seq("ALL", "TRACE", "DEBUG", "INFO", "WARN", "ERROR", "OFF")
+  private val logLevelsString = logLevels.map(x => s"`$x`").reduce((x, y) => s"$x, $y")
```
Zejnilovic: I would be scared someone will write ```--log-level `INFO` ```, would it work?

cerveada: No. The accents were already used in help message, so I reused them in error message as well. Do you think we should drop the accents from both?

migrator-tool/src/main/scala/za/co/absa/spline/migrator/MigratorCLI.scala
cerveada: Make sense, I will change that.

migrator-tool/src/main/scala/za/co/absa/spline/migrator/MigratorCLI.scala
wajda: Makes sense, thx Sasa. I'd remove those back ticks.
Also, to avoid hardcoding the constants you could use something like this:
```scala
val logLevels = classOf[Level].getFields.collect { case f if f.getType == f.getDeclaringClass => f.getName }
```

migrator-tool/src/main/scala/za/co/absa/spline/migrator/MigratorCLI.scala
```diff
@@ -35,6 +35,11 @@ object MigratorCLI extends App {
...
+  private val logLevelInvalidMsg = s"<log-level> should be one of: $logLevelsString"
```
wajda: I would rather inline this constant, just like we inline all the other parser messages. The reason is that it only has a single usage and it's specific to one single parser rule. The same holds for the rest of these constants. Alternatively if you want to reuse, say `logLevels` you could move those vals to the parser rule block where they are used. E.g. 
```scala
{
  val a = ...
  val b = ...
  (opt[...](...))
}
```

migrator-tool/src/main/scala/za/co/absa/spline/migrator/MigratorCLI.scala
Zejnilovic: This is great.  Could this, with some genericity, fit into commons?

migrator-tool/src/main/scala/za/co/absa/spline/migrator/MigratorCLI.scala
```diff
@@ -79,13 +79,15 @@ class MigratorCLI(migratorTool: MigratorTool) {
...
-      val logLevels = classOf[Level].getFields.collect { case f if f.getType == f.getDeclaringClass => f.getName }
-      val logLevelsString = logLevels.reduce((x, y) => s"$x, $y")
-
-      (opt[String]('l', "log-level")
-        text s"Log level ($logLevelsString). Default is ERROR"
-        validate (l => if (logLevels.contains(l)) success else failure(s"<log-level> should be one of: $logLevelsString"))
-        action ((str, conf) => conf.copy(logLevel = Level.valueOf(str))))
+      {
+        val logLevels = classOf[Level].getFields.collect { case f if f.getType == f.getDeclaringClass => f.getName }
+        val logLevelsString = logLevels.reduce((x, y) => s"$x, $y")
```
wajda: could be simplified to just `logLevels.mkString(", ")`

Issue #575, opened 2020-01-30
Upgrade Spline core to Scala 2.12


wajda:
Since there is no dependency on Spark from the Spline core nothing stops us from upgrading to newer a Scala 2.12.
(We cannot use 2.13 as there are a few dependencies that don't support it yet)

PR #576, opened 2020-01-30
spline #575 Upgrade Spline core to Scala 2.12

wajda:


commons/src/main/scala/za/co/absa/spline/common/future/EstimableFuture.scala
```diff
@@ -57,7 +61,7 @@ object EstimableFuture {
         new MovingAverageCalculator(10.seconds.toMillis, 0.05)
       })
       val startTime = System.currentTimeMillis
-      future.onSuccess { case _ => durationMeasurer.addMeasurement(System.currentTimeMillis - startTime) }
+      future.foreach { case _ => durationMeasurer.addMeasurement(System.currentTimeMillis - startTime) }
```
cerveada: What is the reason behind this change?

wajda: `onSuccess` is deprecated in Scala 2.12

persistence/pom.xml
```diff
@@ -62,9 +62,9 @@
             <version>6.5.0</version>
         </dependency>
         <dependency>
-            <groupId>com.arangodb</groupId>
-            <artifactId>velocypack-module-scala</artifactId>
-            <version>1.1.0</version>
+            <groupId>com.github.wajda.com.arangodb</groupId>
+            <artifactId>velocypack-module-scala_${scala.compat.version}</artifactId>
+            <version>1.1.0.1</version>
```
cerveada: It looks like, they merged your changes. https://github.com/arangodb/java-velocypack-module-scala/releases/tag/v1.2.0

So we can use the original version again, right?

wajda: I'm still waiting for it to appear in Maven Central. It's not there yet - https://repo1.maven.org/maven2/com/arangodb/velocypack-module-scala_2.12/1.2.0/velocypack-module-scala_2.12-1.2.0.pom

Issue #442, opened 2019-11-02
Search by attribute name


wajda:
> What kind of information would you like to search?

by column name
Currently in 0.3.9 if we like to know a position of a column in a lineage graph, we have to click through bunch of nodes(projection or aggregation of filter) to find the column and click on the right detail list to look up the lineage in orange. It'd be the best to search by column name then UI can list all possible nodes, then when user click on each search result it would be just like click on the associated node in the graph.

_Originally posted by @iamgzy in https://github.com/AbsaOSS/spline/issues/428#issuecomment-548956567_

cerveada:
Rest API:
INPUT: attribute name (any string)
OUTPUT: attribute ID, execution event ID

The backend should search for all attributes that are on output of any execution plan. For the same write target only the attribute form the latest executed execution should be returned.

Zejnilovic:
When I input `ID1` in search box I get `500` response in UI and `404` on rest server-side saying: 

```
java.util.concurrent.CompletionException: com.arangodb.ArangoDBException: Response: 404, Error: 1203 - AQL: collection or view not found: attributeSearchView (while parsing)
	at java.util.concurrent.CompletableFuture.encodeRelay(CompletableFuture.java:326)
	at java.util.concurrent.CompletableFuture.completeRelay(CompletableFuture.java:338)
	at java.util.concurrent.CompletableFuture.uniRelay(CompletableFuture.java:925)
	at java.util.concurrent.CompletableFuture$UniRelay.tryFire(CompletableFuture.java:913)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
	at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990)
	at com.arangodb.async.internal.velocystream.VstCommunicationAsync.lambda$execute$0(VstCommunicationAsync.java:66)
	at java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:774)
	at java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:750)
	at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
	at java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:1975)
	at com.arangodb.async.internal.velocystream.VstConnectionAsync.lambda$write$0(VstConnectionAsync.java:48)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #611, opened 2020-03-16
spline #442 attribute search back-end

cerveada:


persistence/src/main/scala/za/co/absa/spline/persistence/ArangoInit.scala
```diff
@@ -140,4 +145,35 @@ object ArangoInit extends ArangoInit {
         val functionBody = ARM.using(Source.fromURL(res.getURL))(_.getLines.mkString)
         db.createAqlFunction(functionName, functionBody, new AqlFunctionCreateOptions()).toScala
       }))
+
+  private val searchViewName = "attributeSearchView"
+
+  private def deleteViews(db: ArangoDatabaseAsync) = {
+    val view = db.view(searchViewName)
```
wajda: This line throws an exception if the view doesn't exist.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/AttributeRepositoryImpl.scala
```diff
@@ -0,0 +1,76 @@
...
+
+            LET lastWriteExec = FIRST(FOR exg IN groups SORT exg.timestamp DESC RETURN exg.executionPlan)
+            RETURN {
+              "dataSourceKey": dsk,
+              "executionPlan": lastWriteExec
+            }
+        )
+
+        LET results = (
+          For eps IN executionPlanWithSource
```
wajda: nit: `For` should be in upper case

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/AttributeRepositoryImpl.scala
```diff
@@ -0,0 +1,76 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+
+@Repository
+class AttributeRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends AttributeRepository {
+
+  import za.co.absa.spline.persistence.ArangoImplicits._
+
+  def findAttributesByPrefix(search: String)(implicit ec: ExecutionContext): Future[Array[FoundAttribute]] = {
+    val result = db.queryStream[FoundAttribute](
+      """
```
wajda: Let's use margins (`|`) in the multi-line string. It makes the code easier to read (the boundary between the inner and outer code is clearly visible), and it also makes it easier to handle indents when copy/pasting or auto-formatting.

wajda: nit: indents. The commons style for AQL is 4-space tab.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/AttributeRepositoryImpl.scala
```diff
@@ -0,0 +1,76 @@
...
+                  "id": att.id,
+                  "name": att.name,
+                  "attributeType": type,
+                  "executionEventId": eps.executionPlan._key,
+                  "executionEventName": eps.executionPlan.extra.appName
+                }
+            )
+        )
+
+        FOR r IN FLATTEN(results) RETURN r
```
wajda: When you have nested returns followed with `flatten` it always a sign that the query could be simplified..
E.g.:
```aql
FOR exec IN attributeSearchView
    SEARCH STARTS_WITH(exec.extra.attributes.name, @searchTerm)
    
    LET timestamp = FIRST(FOR prog IN 1 INBOUND exec progressOf SORT prog.timestamp DESC RETURN prog.timestamp)
    LET dataSourceKey = FIRST(FOR ds IN 1 OUTBOUND exec affects RETURN ds._key)
    
    COLLECT dsk = dataSourceKey INTO groups = {
        "executionPlan": exec,
        "timestamp": timestamp
    }
    
    LET lastWriteExec = FIRST(FOR exg IN groups SORT exg.timestamp DESC RETURN exg.executionPlan)
    
    FOR att IN lastWriteExec.extra.attributes
        FILTER CONTAINS(att.name, @searchTerm, true) == 0
        SORT att.name
        LET type = FIRST(FOR t IN lastWriteExec.extra.dataTypes FILTER t.id == att.dataTypeId RETURN t)
        RETURN {
            "id": att.id,
            "name": att.name,
            "attributeType": type,
            "executionEventId": lastWriteExec._key,
            "executionEventName": lastWriteExec.extra.appName
        }
```

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/AttributeRepositoryImpl.scala
```diff
@@ -0,0 +1,76 @@
...
+
+      """, Map("searchTerm" -> search)
+    )
+
+    result.map(_.toArray)
```
wajda: What if there are thousands of results?

persistence/src/main/scala/za/co/absa/spline/persistence/ArangoInit.scala
cerveada: Why do you think so? It does not throw any exceptions for me. It would not make sense to have exists method on the view if it did.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/AttributeRepositoryImpl.scala
cerveada: Yeah, I forgot about this one, lets solve it with AQL LIMIT for now. We can add paging when we implement advanced search.

client-ui/src/app/components/header/attribute-search-bar/attribute-search-bar.component.html
```diff
@@ -0,0 +1,33 @@
...
+</ng-template>
+
+
+<input id="typeahead-input"
+       type="text"
+       class="form-control"
+       placeholder="Search for an attribute..."
+       [ngbTypeahead]="search"
+       [resultTemplate]="rt"
+       [inputFormatter]="onItemSelected"
```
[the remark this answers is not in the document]
cerveada: `ngModel` is triggered every time the text changes. i`nputFormatter` is triggered only when user selects the item.

persistence/src/main/scala/za/co/absa/spline/persistence/ArangoInit.scala
wajda: Hm, you are right. It was throwing for me, but I cannot reproduce it anymore :(
Please ignore this one then.

client-ui/src/app/components/header/attribute-search-bar/attribute-search-bar.component.html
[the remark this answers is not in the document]
wajda: @oleksandr-korchak could you take a quick look at this one please?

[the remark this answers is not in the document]
oleksandr-korchak: @wajda @cerveada  I think, that `inputFormatter` implementation is correct according to the documentation. Of course, there are a lot of questions for the directive authors why did they implemented it in such a way ... but if we want to use it, Adam used it in the right way.

persistence/src/main/scala/za/co/absa/spline/persistence/ArangoInit.scala
```diff
@@ -140,4 +145,35 @@ object ArangoInit extends ArangoInit {
         val functionBody = ARM.using(Source.fromURL(res.getURL))(_.getLines.mkString)
         db.createAqlFunction(functionName, functionBody, new AqlFunctionCreateOptions()).toScala
       }))
+
+  private val searchViewName = "attributeSearchView"
```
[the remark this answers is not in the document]
cerveada: When refactoring I found there is probably bug during db upgrade  https://github.com/arangodb/arangodb-java-driver-async/issues/21

PR #618, opened 2020-03-25
spline #616 Environment variables naming

### LINKED ISSUE #616: Environment variables naming
## Describe the bug
Hi team, thanks a lot for this project! We are trying this out in a production setting :)

Not really a bug but I thought this was the best place anyway. Your compose uses variables like `spline.database.connectionUrl`, directly mapping to the webenv variables. Not really convention-friendly i.e. `VARIABLE`; Among other things, this is also an issue on some provisioners, such as Azure Container Instances - deploying is not possible because the naming convention - [enforced](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-troubleshooting#naming-conventions) by ACI - prevents it. So something like a wrapper var `SPLINE_DATABASE_CONNECTION_URL` could help.

Thanks,

cc @patrickdehoon

wajda:
Fixes #616

### DIFF
--- a/commons/src/main/scala/za/co/absa/spline/common/config/UpperCaseEnvironmentConfiguration.scala
+++ b/commons/src/main/scala/za/co/absa/spline/common/config/UpperCaseEnvironmentConfiguration.scala
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2020 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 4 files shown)


commons/src/main/scala/za/co/absa/spline/common/config/UpperCaseEnvironmentConfiguration.scala
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.common.config
+
+import org.apache.commons.configuration.EnvironmentConfiguration
+import org.apache.commons.lang.StringUtils
+
+class UpperCaseEnvironmentConfiguration extends EnvironmentConfiguration {
```
cerveada:
```suggestion
class NotationInsensitiveEnvironmentConfiguration extends EnvironmentConfiguration {
```

wajda: I have even better idea - spit the responsibility. I can make this class only doing UPPER_SNAKE_NOTATION, and instead of calling the `super` method for the fallback add another `EnvironmentConfiguration` instance to the default config resolver stack

cerveada: `UPPER_SNAKE_NOTATION` interesting 🐍 😄

Issue #615, opened 2020-03-23
Connection Performance Improvements (Timeout config, gzip)


cerveada:
- Allow Timeout configurations for main Spline Componenets
- Use compression for main Spline Componenets

Goal of this ticket is to allow performance related configuration for spline connections and also improve the performance of those connections without refactoring the application too much.

cerveada:
#### Compression
##### Response Compression
Server side compression is pretty standard. It must be switch on servlet container (Tomcat) and the  application running on server doesn't need to care about this. To improve performance there is possibility to do this even sooner on http server (Apache http, Nginx ...).

The client must send `Accept-Encoding` header to let the server know that it is able to deal with encoded response. The server will then compress the content and add `Content-Encoding` header to response.

- Spark agent is already doing this. This is set up by calling `.compress(true)` on scalaj http client.
- Spline UI is also doing this, because in case of Angular app the compression is handled by browser (even for http calls from js). Browser will both set the header and decompress the content.

##### Request Compression
Compression of requests is not that widely supported. This is due to nature of Client server communication where most of the data flow from Server to client. (Unfortunately for spline agent it's exact opposite) Another reason may be so-called zip bomb.

As far as I know there is no support in tomcat, but there probably is some in http servers. To let the server know that the content is compressed the `Content-Encoding` header should be used as in case or response.

To solve this problem I implemented compression in agent and also filter in gateway that will decompress the request body before it reaches the rest controllers.

There is no need for request compression from the ui side, since the requests are expected to be small.

PR #622, opened 2020-04-01
spline #615 request decompression and build info improvements

cerveada:


producer-services/src/main/scala/za/co/absa/spline/producer/service/model/ServerStatus.scala
```diff
@@ -0,0 +1,27 @@
...
+package za.co.absa.spline.producer.service.model
+
+import io.swagger.annotations.ApiModel
+
+
+@ApiModel(description = "Attribute Dependency Node")
+case class ServerStatus(
+  status: String,
+  version: String,
+  requestDecompression: Boolean
```
wajda, line 26: I'm thinking about how the status JSON will look like when and if we add more flags to it.
Although any string would work, ideally we need to have some naming convention to follow.

First of all I would name a feature not after what it's doing under the hood, but rather after what I can use it for.
E.g. as a client I want to know if I can send a compressed request to a server, so for the respective feature flag name, `supportsRequestCompression` would do better than `requestDecompression` IMO.

The naming convention I'd like to propose is: `supportsXXXX: true` for features.
Another approach would be to have a `flags:[]` properties that would simply list all features enabled. E.g.
```json
{
  "version": "1.2.3",
  "flags": [ "request-compression", "foo", "bar" ]
}
```

What do you think?

producer-rest-core/src/main/scala/za/co/absa/spline/producer/rest/controller/StatusController.scala
```diff
@@ -42,10 +44,24 @@ class StatusController @Autowired()(
...
+  @RequestMapping(
+    path = Array("/status"),
+    method = Array(RequestMethod.GET))
+  @ApiOperation(
+    value = "Server status",
+    notes = "Returns server status, version and abilities")
+  def statusGet(response: `HttpServletResponse`): Future[ServerStatus] = repo
+    .isDatabaseOk
+    .map({
+      if (_) "OK"
```
wajda, line 63: `GET` and `HEAD` should have the same contract in terms of response codes and headers.
See: https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.4

Issue #612, opened 2020-03-19
Lineage Timestamp format is missleading


Zejnilovic:
## Describe the bug
Timestamp in the Lineage overview is misleading. Shows only hours up to 12 and no PM/AM

## Components State
- [x] ArangoDB running without errors
- [x] ArangoDB spline database initialized
- Rest Gateway running and
  - [x] connects to ArangoDB
  - [x] there are no errors in logs
- Spline UI running and
  - [x] connects to Rest Gateway consumer
  - [x] there are no errors in logs

## To Reproduce
1. Go to spline UI
2. Pick any lineage and open it
3. Look at the Left Panel timestamp

## Expected behaviour
I would expect to either have a 24h format or a PM/AM at the end.

#### This job was run at 17:10 or 05:10 PM
![image](https://user-images.githubusercontent.com/17467768/77041854-ea106d00-69ba-11ea-8168-41bba073520e.png)

wajda:
Confirm. UI is using a strange time format. Should be 24h everywhere.

Zejnilovic:
> Confirm. UI is using a strange time format. Should be 24h everywhere.

In that case. Dashboard need an update too
![image](https://user-images.githubusercontent.com/17467768/77064520-e93e0200-69df-11ea-86ec-aaab8ad376bf.png)

Don't know of any other places.

Zejnilovic:
Works as expected

PR #627, opened 2020-04-08
#612 :: 24h time format everywhere & fix prod build & auto open brows…

oleksandr-korchak:


client-ui/src/app/components/lineage/lineage-details/operation-properties-details/properties/properties.component.ts
```diff
@@ -41,7 +41,9 @@ export class PropertiesComponent implements OnDestroy {
   public propertyName: string
 
   @Input()
-  public properties: Property[]
+  public properties: Property[];
+
+  PropertyType = PropertyType;
```
wajda, line 46: nit: we don't use semicolons in JS/TS code

oleksandr-korchak: So, you are doing wrong. This is not Python and not JS, but Typescript. Have a look at any project, Angular, Angular CDK, React, Vu.js source code and you will see semicolumns everywhere.

oleksandr-korchak: Here is what Sonar source say:

> 
**Statements should end with semicolons**
In JavaScript, the semicolon (;) is optional as a statement separator, but omitting semicolons can be confusing, and lead to unexpected results because a semicolon is implicitly inserted at the end of each line.

>

wajda: There are long-long-long and never ending debates about automatic semicolon inference and its pros and cons. Simply saying that ins't _wrong_ just because Angular or Vue folks are doing it in another way is not wise. We could have have this discussion later and we might change it in the future, but for now consistency is more important. Let's discuss it offline.

Issue #613, opened 2020-03-20
Prevent getting out of displayed lineage in embeddedMode


benedeki:
## Background
When Spline 0.4.x UI is displayed in embedded mode (`embeddedMode = true`), it's possible to navigate out of the requested lineage.

## Feature
Disable the "clickability" of components leading to overview or completely out of Spline, if `embeddedMode = true`.

## Example
![image](https://user-images.githubusercontent.com/14905969/77151695-c9b5e080-6a96-11ea-84a7-d78d611a2843.png)

wajda:
We'll put a quick fix to the 0.5 release, but the 'embeddedMode` as it is, should be revised in #614

benedeki:
0,5 release or 0.4, please?

wajda:
let's discuss on Monday

Zejnilovic:
Look good in Menas

PR #628, opened 2020-04-08
#613 :: Hide header in embeddedMode.

oleksandr-korchak:


client-ui/src/app/app.component.html
```diff
@@ -15,6 +15,8 @@
   -->
 
 
-<app-header></app-header>
+<app-header *ngIf="(isEmbeddedMode$ | async) === false"></app-header>
```
wajda, line 18: Why not simply `*ngIf="!(isEmbeddedMode$ | async)"`?

client-ui/src/app/app.component.ts
```diff
@@ -15,12 +15,21 @@
...
+    selector: 'app-root',
+    templateUrl: './app.component.html',
+    styleUrls: ['./app.component.less']
 })
 export class AppComponent {
 
+    isEmbeddedMode$: Observable<boolean>;
+
+    constructor(private readonly store: Store<AppState>) {
+        this.isEmbeddedMode$ = this.store.select('config', 'embeddedMode');
```
wajda, line 33: nit: semicolon

client-ui/src/app/app.component.html
oleksandr-korchak: Because it is easy to make a mistake and to miss "!".

That is why when evaluating:

true:

`*ngIf="isEmbeddedMode$ | async"`

false:

`*ngIf="(isEmbeddedMode$ | async) === false"`

PR #621, opened 2020-03-26
SSL support in the admin client - draft 

rvvincelli:
See [here](https://github.com/AbsaOSS/spline/issues/620) for the feature request.

A new schema `arangodbs` is added.  The SSL implementation is unsecure - the server certificate is not validated. Because of this, the user must explicitly give a `-k` `--insecure` switch to allow for an SSL connection.

Tests are (not) yet implemented - I might have some time later today or in the coming days; but you have time too team, even better maybe.

Thanks,

cc @wajda @patrickdehoon

Issue #620, opened 2020-03-25
SSL support


rvvincelli:
## Background
Hi team, thanks for quickly addressing my other issue today.

First of all: should this issue be a duplicate, my apologies - just link me to the first one; also because SSL support sounds important - perhaps I'm missing something myself; it's been a long coding day - sorry in advance.

As we make progress, the ArangoDB instance we use works on HTTPS because it is a production one.

When initializing the db with:
`java -jar admin-0.4.2.jar db-init http+ssl://root:test@<IP>/spline`
we get:
`java.io.IOException: Reached the end of the stream.`
This version is the latest I see.

We know that this could be caused by the protocol mismatch, but the command works against an unsecure Arango, and we have the suspect that this is just a failed SSL handshake actually. With the official `arangosh` client, making an unsecure connection to a secure server will fail indeed.

By looking at the code it looks like the builder has no SSL at all, see [here](https://github.com/AbsaOSS/spline/blob/master/persistence/src/main/scala/za/co/absa/spline/persistence/ArangoDatabaseFacade.scala#L28), indeed there is only one schema in the admin client, `arangodb`. In the Java driver they have an [example](https://github.com/arangodb/arangodb-java-driver/blob/master/src/test/java/com/arangodb/example/ssl/SslExample.java#L71).

## Feature
The requested feature is to support SSL, possibly with a different scheme, like `arangodbs` or `arangodb+ssl`.

## Example [Optional]
See the builder example for SSL above from the driver codebase.

## Proposed Solution [Optional]
It does not seem like a lot of work so I cannot promise anything but we might submit a patch ourselves - that would take time though.

Thanks again!

cc @patrickdehoon

wajda:
> When initializing the db with:
java -jar admin-0.4.2.jar db-init http+ssl://root:test@<IP>/spline
we get:
java.io.IOException: Reached the end of the stream.

That's strange, you should get `java.net.MalformedURLException` instead. At the moment Spline only support ArangoDB connection string starting with `arangodb://`

Regarding the rest I agree. Will see if we can squeeze it into 0.5 release

wajda:
I would propose to change the connection URL schema. Instead of `arangodb` we could use the protocol that will be actually used for the connection, e.g:
```
# unsecured
vst://
http+json://
http+vpack://

# secured
vst+ssl://
https+json://
https+vpack://
```

For backward compatibility, we can also leave `arangodb://`, that would mean a default protocol (currently `vst`).

[...]

PR #632, opened 2020-04-08
Spline #620 - Adds SSL support in the admin client

wajda:
This is a re-based #621 + some minor fixes

Thanks @rvvincelli


rvvincelli:
Thanks to you and team @wajda for this project! We are going forward with it ;)

PR #658, opened 2020-04-24
Bugfix/spline 657 attr search navigation

wajda:


client-ui/src/app/components/lineage/lineage-graph/lineage-graph.component.ts
```diff
@@ -58,25 +57,13 @@ export class LineageGraphComponent implements OnInit, OnChanges, AfterViewInit {
...
   private cytograph: CytoscapeNgLibComponent
 
-  public ngOnInit(): void {
-    const writeNode = this.graph.nodes.find(n => n.data._type == 'Write')
-    writeNode.data.icon = operationIconCodes.get(OperationType.Write)
-    writeNode.data.color = operationColorCodes.get(OperationType.Write)
-  }
-
   public ngOnChanges(changes: SimpleChanges): void {
+    if (changes['graph']) this.refreshGraph()
```
oleksandr-korchak, line 61: I guess, you do not want to do that on the very first change as far aa you are doing that in th ngAfterInit then one more time. 

So it should be:

`
if (changes['graph'] && !changes['graph'].isFirstChange()) {
      this.refreshGraph()
    }
`

client-ui/src/app/components/lineage/lineage-graph/lineage-graph.component.ts
```diff
@@ -87,10 +74,28 @@ export class LineageGraphComponent implements OnInit, OnChanges, AfterViewInit {
         this.selectedNodeChange.emit(nodeId)
       })
     })
+    this.refreshGraph()
     this.refreshSelectedNode()
     this.refreshAttributeGraph()
   }
 
+  public refreshGraph(): void {
```
oleksandr-korchak, line 82: Why it should be public?

client-ui/src/app/components/lineage/lineage-graph/lineage-graph.component.ts
wajda: So I guess I need to do it for all three `if`s

client-ui/src/app/components/lineage/lineage-graph/lineage-graph.component.ts
wajda: a typo. This used to be a `ngAfterViewInit` before.
Thanks for pointing out.
Fixed.

PR #660, opened 2020-04-27
Bugfix/spline 659 disable attr search on 03 data

wajda:


client-ui/src/app/components/lineage/lineage-graph/lineage-graph.component.ts
```diff
@@ -126,13 +126,11 @@ export class LineageGraphComponent implements OnChanges, AfterViewInit {
     })
   }
 
-  private highlightAttrLinAndImp(attrLinGraph: AttributeGraph, attrImpGraph: AttributeGraph) {
+  private highlightAttrLinAndImp(attrLinGraph: AttributeGraph | undefined, attrImpGraph: AttributeGraph) {
```
oleksandr-korchak, line 129: if the second argument cannot be undefined, then the second argument should be first:

`
highlightAttrLinAndImp(attrImpGraph: AttributeGraph, attrLinGraph?: AttributeGraph)
`

client-ui/src/app/components/lineage/lineage-highlights-toolbar/lineage-highlights-toolbar.component.html
```diff
@@ -16,46 +16,58 @@
...
+      </div>
+
+      <mat-icon fontSet="fa"
+                fontIcon="fa-info-circle"
+                [matTooltip]="legend.description">
+      </mat-icon>
     </div>
-    <!-- ./LEGEND -->
+  </div>
+
```
oleksandr-korchak, line 59: You are missing here closing comment

client-ui/src/app/components/lineage/lineage-highlights-toolbar/lineage-highlights-toolbar.component.less
```diff
@@ -61,4 +61,13 @@
     color: #ccc;
   }
 
+  .old-spline-warn {
+    margin-top: 15px;
```
oleksandr-korchak, line 65: All definitions of margins & paddings should be devided by rem (which is 16px) in that box. It means, it can be 4px, 8px, 12px, 16px, 24px etc. wchi is 0.25rem, 0.5rem, 0.75rem, 1rem, 1.5rem,

client-ui/src/app/components/lineage/lineage.component.ts
```diff
@@ -61,13 +65,25 @@ export class LineageComponent implements OnDestroy {
       .select('router', 'state', 'queryParams', 'selectedNode')
       .subscribe((nodeId: string) => {
         this.selectedNodeId = nodeId
+        console.log("NODE", nodeId)
```
oleksandr-korchak, line 68: remove debugging stuff.

client-ui/src/app/components/lineage/lineage.component.ts
```diff
@@ -61,13 +65,25 @@ export class LineageComponent implements OnDestroy {
...
         )
       })
     )
 
+    this.subscriptions.push(this.store
+      .select('attributeLineageAndImpact').pipe(filter(_.identity))
+      .subscribe(({impact}: AttributeLineageAndImpact) => {
+        if (!this.selectedNodeId) {
+          const primaryAttr = getImpactRootAttributeNode(impact)
+          console.log("IMPACT")
```
oleksandr-korchak, line 81: remove it.

client-ui/src/app/components/lineage/lineage.component.html
```diff
@@ -18,15 +18,17 @@
...
                      [attributeLineageAndImpactGraph]="d.attributeLinAndImp"
                      [selectedNode]="selectedNodeId"
                      (selectedNodeChange)="onNodeSelected($event)">
       </lineage-graph>
 
       <lineage-highlights-toolbar
               *ngIf="selectedAttribute$ | async as selectedAttributes"
+              [lineageGraphLegends]="lineageGraphLegendsToShow$ | async"
               [attributes]="[selectedAttributes]"
+              [showOldSplineWarning]="isExecPlanFromOldSpline(d.plan.executionPlan)"
```
oleksandr-korchak, line 31: do not use function in the template. You should calculate it in a different way.

client-ui/src/app/model/lineage-graph/leneage-graph-legend.models.ts
```diff
@@ -15,26 +15,31 @@
  */
 
 
-import { LINE_COLOR_HLT_IMPACT, LINE_COLOR_HLT_LINEAGE, LINE_COLOR_HLT_PRIMARY } from './leneage-graph.models';
+import {LINE_COLOR_HLT_IMPACT, LINE_COLOR_HLT_LINEAGE, LINE_COLOR_HLT_PRIMARY} from './leneage-graph.models';
 
 
-export interface LineageGraphLegend {
+export type LineageGraphLegend = {
```
oleksandr-korchak, line 21: Why it should be a type over the rule "interface-over-type-literal"?

client-ui/src/app/model/lineage-graph/leneage-graph-legend.models.ts
```diff
@@ -15,26 +15,31 @@
...
-  return [
-    {
+export type LineageGraphLegendKeys =
+  "Usage" |
+  "Lineage" |
+  "Impact"
+
+export const LineageGraphLegends: { [key in LineageGraphLegendKeys]: LineageGraphLegend } =
+  Object.freeze({
+    Usage: {
```
oleksandr-korchak, line 34: What will happen with that field if you will decide to refactor LineageGraphLegendKeys and rename Usage to somethoing else i ncase of having LineageGraphLegendKeys as a type?

client-ui/src/app/model/lineage-graph/leneage-graph-legend.models.ts
```diff
@@ -15,26 +15,31 @@
...
+export type LineageGraphLegend = {
   title: string;
   color: string;
   description: string;
 }
 
-export function getLineageGraphLegend(): LineageGraphLegend[] {
-  return [
-    {
+export type LineageGraphLegendKeys =
```
oleksandr-korchak, line 27: It is better to use enum here like so:

```
export enum LineageGraphLegendType {
  Usage = 'usage',
  Lineage = 'lineage',
  Impact = 'impact'
}
```

and then use it in object definition and in all app to access that:

```
LineageGraphLegends[LineageGraphLegendType.Usage]
```

client-ui/src/app/model/lineage-graph/leneage-graph-legend.models.ts
```diff
@@ -15,26 +15,31 @@
...
+  "Usage" |
+  "Lineage" |
+  "Impact"
+
+export const LineageGraphLegends: { [key in LineageGraphLegendKeys]: LineageGraphLegend } =
```
oleksandr-korchak, line 32: We did not disscuss nameing convention yet, but still the current name format of the const means for me that this is a class or interface. Global const definition is UPPER_SNACK_CASE.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/internal/AttributeDependencySolver.scala
```diff
@@ -25,19 +25,24 @@ import za.co.absa.spline.consumer.service.internal.model.ExecutionPlanDAG
...
+    val impact = getAttributeImpact(originOp, attributeId, dependencyResolver)
+    asGraph(impact)
+  }
 
-  def lineage(attributeId: AttributeId): Option[AttributeGraph] =
-    findOriginOperationForAttr(attributeId).map(originOp => asGraph(getAttributeLineage(originOp, attributeId)))
+  def lineage(attributeId: AttributeId, dependencyResolver: AttributeDependencyResolver): AttributeGraph = {
+    val originOp = originOperationForAttr(attributeId)
+    val impact = getAttributeLineage(originOp, attributeId, dependencyResolver)
+    asGraph(impact)
```
cerveada, line 39:
```suggestion
    val lineage = getAttributeLineage(originOp, attributeId, dependencyResolver)
    asGraph(lineage)
```

client-ui/src/app/model/lineage-graph/leneage-graph-legend.models.ts
wajda: I'm pretty sure this rule should be considered outdated. There are many thoughts and debates about types vs interfaces, why both exist in the same language and what to use when, in most cases it's the matter of taste (unfortunately). But I'm in the campus that say that it's the matter of paradigm in use, rather then just taste.
In modern TS there aren't many things that interfaces could do and types are not and vice versa (event in fact the types could do much more things over interfaces then the opposite). 

The ___most important___ difference however is not what they can or cannot do, but what information (the purpose) they communicate to the developer. 

The interfaces as well as classes originate in the Object-Oriented paradigm. Interfaces describe the object contract and are meant to be implemented by some class. Even though it's possible to describe the data structure with an interface, the main focus is usually made on behavioral aspect - the methods.

The types on the other hand is a purely functional  beast. It is made to describe data types in all possible aspects. Types cover any types, not only object ones and are very flexible.

Do the general rule of thumb as I see it is simple - if you aim to implement X in some class in your project and/or X is logically more about behavior then structure then X is an interface. Otherwise X is a type.

client-ui/src/app/model/lineage-graph/leneage-graph-legend.models.ts
wajda: It will stop compiling. This is the intention :) With `[key in LineageGraphLegendKeys]` you clearly define the property names allowed for the object. Basically `enum` semantics.

client-ui/src/app/model/lineage-graph/leneage-graph-legend.models.ts
wajda: IMO this way in comparison to the types approach you are missing the type safely.

client-ui/src/app/model/lineage-graph/leneage-graph-legend.models.ts
wajda: Yeah, this is a CQC topic. Let's leave it for now until we make a statement.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/internal/AttributeDependencySolver.scala
wajda: nice one :)

client-ui/src/app/components/lineage/lineage.component.html
wajda: generally agree, but leaving it as is for now to speed the process up. In this case it's not that important as the function is pure and trivial.

PR #693, opened 2020-06-04
Improve docker compose

DaimonPl:
I had several problems with original compose:

`spline` was not actively waiting for `arangodb` - because of that spline initialization failed in random way
`spline` was not performing check if db is already initialized - because of that it was not possible to run it more than once unless previous instances were removed or compose.yml modified
`arangodb` was not exposed on localhost to access its ui

here are the changes:
 - arangodb now exposes port
 - spline is actively waiting for arangodb before start (using simple curl)
 - spline is performing db initialization only if marker file does not exist


docker-compose.yml
```diff
@@ -3,25 +3,39 @@ version: '3'
...
     image: arangodb:3.5.1
+    restart: always
+    ports:
+      - 8529:8529
     environment:
       ARANGO_NO_AUTH: 1
 
   spline:
-    image: absaoss/spline-rest-server
+    image: "absaoss/spline-rest-server:0.5.1"
```
wajda, line 13: Is it possible to use a variable instead of a hardcoded version?
Something like `absaoss/spline-rest-server:${spline_version}`

docker-compose.yml
```diff
@@ -3,25 +3,39 @@ version: '3'
...
-      && curl -O -s https://repo1.maven.org/maven2/za/co/absa/spline/admin/0.5.0/admin-0.5.0.jar
-      && java -jar ./admin-0.5.0.jar db-init arangodb://arangodb/spline -s
-      && catalina.sh run"
+      bash -c "
+        until curl --output /dev/null --silent --get --fail http://172.17.0.1:8529
+        do
+          echo waiting for arango db
+          sleep 5
+        done
+        if [ ! -f ./init_completed ]
```
wajda, line 24: Why do you need a marker file? To save on startup time?

docker-compose.yml
DaimonPl: it's possible to use environment variables but AFAIK there's no way to define default values for them

so instead of
`docker-compose up`
everyone would need to call
`spline_version=0.5.1 docker-compose up` 
not sure if it's a way to go ?

docker-compose.yml
DaimonPl: nope, you simply get "db already exists" if run twice :( it's not cleaniest, but easiest

wajda: `-s` option tells `admin-cli` to skip initialization if the db is already initialized. Aren't we doing the same thing twice with the marker file?

docker-compose.yml
wajda: What about `.env`-file?
https://docs.docker.com/compose/environment-variables/#the-env-file

DaimonPl: yep that's an option I'll update it

docker-compose.yml
DaimonPl: I'll recheck again but i'm 99% sure that it wasn't working

DaimonPl: Yep

```
spline_1    | Exception in thread "main" java.lang.IllegalArgumentException: Arango Database spline already exists
spline_1    |   at za.co.absa.spline.persistence.ArangoInit$.$anonfun$deleteDbIfRequested$1(ArangoInit.scala:109)
spline_1    |   at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:307)
spline_1    |   at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:41)
spline_1    |   at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
spline_1    |   at java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1402)
spline_1    |   at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
spline_1    |   at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
spline_1    |   at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
spline_1    |   at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
spline_spline_1 exited with code 1

```

wajda: Aha, it means that `admin-cli` has a bug.

wajda: Looks like an ArrangoDB driver bug - fileed an issue #696

wajda: Do you agree to remove the marker file if `-s` option works correctly? I'll place a quick workaround in 0.5.2

DaimonPl: Sure, that problem is the only reason this marker workaround has been added :)

wajda: PR created - #698

docker-compose.yml
DaimonPl: Done

docker-compose.yml
DaimonPl: I've reverted 'init_completed' part

Issue #695, opened 2020-06-04
Improve docker compose


wajda:
_Originally posted by @DaimonPl in 
https://github.com/AbsaOSS/spline/pull/693#issue-427673433_
> I had several problems with original compose:
> 
> `spline` was not actively waiting for `arangodb` - because of that spline initialization failed in random way
> `spline` was not performing check if db is already initialized - because of that it was not possible to run it more than once unless previous instances were removed or compose.yml modified
> `arangodb` was not exposed on localhost to access its ui
> 
> here are the changes:
> 
> * arangodb now exposes port
> * spline is actively waiting for arangodb before start (using simple curl)
> * spline is performing db initialization only if marker file does not exist

DaimonPl:
I think this can be closed, PR is merged

omennemo:
The docker-compose that is in the guide on GH pages is not the same as the updated one. The documentation probably needs an update as well

PR #709, opened 2020-06-08
#695 Ensure startup does not block on entropy availability

DaimonPl:
I had several situations when tomcat would not start for several minutes. I checked logs and it was because of lack of entropy for random generator, so:

 - change to non-blocking entropy provider
 - additionally update and make arangodb version configurable
 - use status endpoint when waiting for arango db


wajda:
LGTM & Thanks.

PS: We have a requirement for each PR to be assigned to an issue. So it would be better if you create an issue first next time, before a PR. It would be also easier from the version planning perspective. For example if the fix can be included into an upcoming hotfix release then the PR has be based on the respective `release/` branch.

Issue #651, opened 2020-04-22
Transformations view enhancment


Zejnilovic:
## Background
When I select a transformation I get a modal window with the transformation but it has no header and the expand buttons are just boxes
![image](https://user-images.githubusercontent.com/17467768/79996774-1d15b700-84b9-11ea-89fb-8d2acedc35ef.png)

## Feature
- Add a header to the transformation detail
- Switch the expand buttons for something like a down-facing arrow and right-facing arrow

wajda:
![image](https://user-images.githubusercontent.com/795479/84424175-fd308300-ac1f-11ea-967b-ae8a3ab5fb89.png)

PR #724, opened 2020-06-11
spline #651 Expression modal view enhancement

wajda:


client-ui/src/app/components/modal/modal-expression/modal-expression.component.less
```diff
@@ -14,76 +14,160 @@
...
+
+              &::before {
+                -moz-osx-font-smoothing: grayscale;
+                -webkit-font-smoothing: antialiased;
+                display: inline-block;
+                font-style: normal;
+                font-variant: normal;
+                text-rendering: auto;
+                line-height: 1;
+                font-family: "Font Awesome 5 Free";
```
oleksandr-korchak, line 150: You will have here an error. There should be at leas one generic font defined, e.g. like so: font-family: "Font Awesome 5 Free", sans-serif;

wajda: What's the point? This is a copy from the FontAwesome CSS.

oleksandr-korchak: The point is that there is a rule in Sonar, which requires definition of some default font which will be used in case Font Awesome font will not be found. 

Agree, it makes no sense in that case. But you have to fix it or disable that rule in Sonar.

wajda: I think we'll rather deal with Sonar. The code clarity and understandability is more important than silly rules.

wajda:
Unfortunately static code analysis suffers from many corner cases, like in this PR. The Sonar complains about CSS selectors that belong to 3rd party library. I cannot do much about it.

Issue #708, opened 2020-06-08
Expression view in Generic operations


wajda:
For the sake of UI consistency and user-friendliness we need expression type properties of _Generic_ operations to be rendered in the same way as it is done in, e.g. Filter, Sort, Aggregate and other operations that can accept expressions as their parameter. 

So, instead of:

```json
  {
    "_typeHint": "expr.Literal",
    "value": 100,
    "dataTypeId": "4ccf49cc-47c8-4665-8bb2-eb172b946ea3"
  },
```

there should be:

![image](https://user-images.githubusercontent.com/795479/84030705-c430ae00-a994-11ea-91b4-790af6c30dd1.png)

PR #725, opened 2020-06-12
#708 :: Operation Property :: expresion property view.

oleksandr-korchak:


client-ui/src/assets/config.json
```diff
@@ -1,4 +1,4 @@
 {
-    "apiUrl": "http://localhost:8080/consumer",
+    "apiUrl": "https://opensource.bigusdatus.com/spline/rest/consumer",
```
wajda, line 2: Is this an accidental commit?

PR #729, opened 2020-06-13
Bugfix/more ui fixes

wajda:


client-ui/src/app/components/header/header.component.ts
```diff
@@ -32,6 +34,8 @@ export class HeaderComponent {
     private store: Store<AppState>
   ) { }
 
+  appVersion: string = __APP_VERSION__
```
oleksandr-korchak, line 37: type definition here is redundant + it should be readonly.

client-ui/src/app/components/header/header.component.html
```diff
@@ -23,6 +23,7 @@
     <div class="collapse navbar-collapse" id="navbarSupportedContent">
         <div class="title">
             Data Lineage Tracking And Visualization
+            <span class="spline-version">|&nbsp;&nbsp;Version {{appVersion}}</span>
```
oleksandr-korchak, line 26: Header title will become a very long string in that way. It is better to display it as a second string or in another place.

client-ui/src/app/components/lineage/lineage-details/operation-properties-details/extra-properties/extra-properties-expression.component.ts
```diff
@@ -25,10 +25,10 @@ export class ExtraPropertiesExpressionComponent {
 
   @Input() properties: OperationProperty.ExtraPropertyValueExpression[]
 
-  @Output() valueClicked$ = new EventEmitter<{ property: OperationProperty.ExtraPropertyValueExpression }>()
+  @Output() expressionClicked$ = new EventEmitter<OperationProperty.ExpressionValue>()
```
oleksandr-korchak, line 28: Why did you refactor EventEmitter payload structure?

client-ui/src/app/model/operation/operation-property.models.ts
```diff
@@ -42,77 +42,61 @@ export namespace OperationProperty {
...
-        extraProperties
+        {
+          primitive: [],
+          expression: [],
+          json: []
+        }
       )
+  }
 
+  function expressionValue(key: string, exprs: Record<string, any>[], attrs: AttributeVM[]): ExtraPropertyValueExpression {
```
oleksandr-korchak, line 79: I would call it as toExpressionValue or extractExpressionValue

client-ui/src/app/model/operation/operation-property.models.ts
```diff
@@ -42,77 +42,61 @@ export namespace OperationProperty {
...
+    return {
+      label: humanizeCase(key),
+      value: exprs.map(e => ({
+        value: getText(e, attrs),
+        rawValue: e
+      }))
+    }
   }
 
+  function humanizeCase(str: string): string {
```
oleksandr-korchak, line 89: => toHumanizeCase

client-ui/src/app/model/operation/operation-property.models.ts
```diff
@@ -42,77 +42,61 @@ export namespace OperationProperty {
...
-
-          if (value === undefined || value === null) {
-            return result
+        (acc, [key, value]) => {
+          if (typeof value === 'string' || typeof value === 'number' || typeof value === 'boolean') {
+            const primitiveValue: ExtraPropertyValuePrimitive = {
+              label: key,
+              value: value
+            }
+            return {...acc, primitive: [...acc.primitive, primitiveValue]}
```
oleksandr-korchak, line 55: you sohuld not be copying the whole object for just push another array  item here. 

acc.primitive.push(primitiveValue)

client-ui/src/app/model/operation/operation-property.models.ts
```diff
@@ -42,77 +42,61 @@ export namespace OperationProperty {
...
           }
-
-          const primitivePropValue = {
-            label: key,
-            value
+          else if (typeof value === 'object' && isExpresionProperty(value)) {
+            return {...acc, expression: [...acc.expression, expressionValue(key, [value], attributesList)]}
```
oleksandr-korchak, line 58: the same issue as with primitive

client-ui/src/app/model/operation/operation-property.models.ts
```diff
@@ -42,77 +42,61 @@ export namespace OperationProperty {
...
           }
-
-          if (typeof value === 'string' || typeof value === 'number') {
-            extraProperties.primitive.push(primitivePropValue)
+          else if (typeof Array.isArray(value) && value.length && isExpresionProperty(value[0])) {
+            return {...acc, expression: [...acc.expression, expressionValue(key, value, attributesList)]}
```
oleksandr-korchak, line 61: the same issue as with primitive

client-ui/src/app/model/operation/operation-property.models.ts
```diff
@@ -42,77 +42,61 @@ export namespace OperationProperty {
...
-              const decoratedValue = decorateExpressionProperty(primitivePropValue, attributesList)
-              extraProperties.expression.push(decoratedValue)
-            } catch (e) {
-              console.error(`Expresion decoration error`, e)
+          else {
+            const jsonValue: ExtraPropertyValueJson = {
+              label: humanizeCase(key),
+              value
             }
+            return {...acc, json: [...acc.json, jsonValue]}
```
oleksandr-korchak, line 68: the same issue as with primitive

client-ui/src/app/model/operation/operation-property.models.ts
```diff
@@ -42,77 +42,61 @@ export namespace OperationProperty {
...
   export function parseExtraOptions(nativeProperties: NativeProperties, attributesList: AttributeVM[]): ExtraProperties {
-    const extraProperties: ExtraProperties = {
-      primitive: [],
-      expression: [],
-      json: []
-    }
-
-    return Object.keys(nativeProperties)
+    return Object.entries(nativeProperties)
+      .filter(([, v]) => v != null)
```
oleksandr-korchak, line 47: having here filter instead of if statement inside of reduce you are iterating 2 times over whole array instead of 1. I don't think that this is a better approach.

client-ui/src/app/components/lineage/lineage-details/operation-properties-details/extra-properties/extra-properties-expression.component.ts
oleksandr-korchak: Using an object as EventEmitter payload should be preferred over pure value, even if it is the only one value there so far.

By doing so you are prepared for easy payload extension and also aby calling object property in a proper way you are doing life easier with working with payload in other places.

client-ui/src/app/components/modal/modal-expression/modal-expression.component.less
```diff
@@ -150,6 +149,25 @@
                 font-family: "Font Awesome 5 Free";
                 content: "\f02b";
                 color: darkcyan;
+                font-size: 12px;
+              }
+            }
+
+            &.expr-Alias {
```
oleksandr-korchak, line 156: please, use dash-case all lower naming convention for CSS classes names. 

expr-Alias => expr-alias

client-ui/src/app/model/operation/operation-property.models.ts
wajda: "humanize" is a verb

client-ui/src/app/model/operation/operation-property.models.ts
wajda: I would agree if the algorithm were mutable. But in this case `reduce` function is used, which implies immutability. And modifying an accumulator from inside a reducer is quite a mistake.

client-ui/src/app/model/operation/operation-property.models.ts
wajda: This is how you keep you reducer smaller and more focused it's primary function. This algorithm is dealing with usually very short arrays, so the code clarity should be a priority over tiny hypothetical speed gain.

client-ui/src/app/components/modal/modal-expression/modal-expression.component.less
wajda: In this case `Alias` is a dynamic part of a style class name, which is taken from the `Expression._typeHint`. Ideally I would prefer to leave the original value (i.e. `expr.Alias`), but since dot is a reserved character CSS I had to escape it. `\.` notation doesn't seem to be a standard, so I replaced it with a dash doesn't belong to the value domain and couldn't create collisions. Yes I could replace camel case with a sneak case and lowercase it all, but the more changes you do to the value the more difficult it gets to look it up and distinguish later. 
That's why I opted to only do a single mandatory replacement `.`->`-`

client-ui/src/app/components/header/header.component.ts
wajda: +1 for `readonly`.
For type definition, technically it is optional, but usually type defs helps a lot to reading if the type isn't obvious from the context. Also we require type defs for public members, and `appVersion` is a public member.

client-ui/src/app/components/header/header.component.html
wajda: Are you volunteering? :)

client-ui/src/app/components/lineage/lineage-details/operation-properties-details/extra-properties/extra-properties-expression.component.ts
wajda: Good point for readability in the usage place, e.g. `$event.name` versus `$event.property.name`. Thanks.

client-ui/src/app/model/operation/operation-property.models.ts
oleksandr-korchak: Oh, you are right. In that case it is ok.

client-ui/src/app/model/operation/operation-property.models.ts
oleksandr-korchak: Why the algorithm should be immutable in that case? There is no need to use more memory and make code less readable here. 

What kind of mistake you are talking about? I cannot see any mistake by working with one object all the time inside a reduce function and just push new values to the array in the current case.

client-ui/src/app/components/header/header.component.html
oleksandr-korchak: Should I? 
According to me, it was good enough before. I do not see any reason to solve it at all so far. Isn't it better to leave it as it was before?

wajda: No, it wasn't good enough
image

client-ui/src/app/model/operation/operation-property.models.ts
wajda: We prefer functional code style for everything unless there are reasons to do opposite.

wajda: > I cannot see any mistake by working with one object all the time inside a reduce function and just push new values to the array in the current case.

Of cause, technically it's not an issue. The compiler would be happy. The problem is laying in violation one of core FP principles - _breaking immutability_.  It is extremely important from the perspective or code reading and maintenance. In the above example by mutating `acc` instance from inside a reducer function you defeat the purpose of an accumulator, and effectively turn `reduce` function into `forEach`-ish. The resulting code however doesn't resemble neither of those, and raises too many "why"s to a reader. 

Sure, paradigms can be mixed in practical coding, but it should be done with care, in a cohesive manner, avoiding collisions. After all, programming paradigms exist to decrease the level of entropy in software development - their main purpose is to create restriction rather than giving new possibilities. So it's very important to understand and obey their core principles.

client-ui/src/app/components/lineage/lineage-details/operation-properties-details/extra-properties/extra-properties-expression.component.ts
wajda: Although I like the suggestion in general, in this particular case however, I don't think it adds any benefits. The payload itself is already a wrapper with quite opaque semantics:
```typescript
export type ExpressionValue = {
    value: string
    rawValue: Record<string, any>
  }
``` 
It I create another wrapper on a wrapper that would look strange. I even struggle to create a meaningful name for it. E.g.
```typescript
// definition 
new EventEmitter<{ value: OperationProperty.ExpressionValue }>
// usage
"$event.value.value" // Doh...
```

client-ui/src/app/components/header/header.component.html
oleksandr-korchak: I cannot see that happening on full HD resolution. So you are solving here some edge case. And it is good enough if you are supporting desktop so far.

client-ui/src/app/components/lineage/lineage-details/operation-properties-details/extra-properties/extra-properties-expression.component.ts
oleksandr-korchak: it will be even worse if you would call as X, like so:

`
new EventEmitter<{ x: OperationProperty.ExpressionValue }>
`

but my point is to call it in a way it will help you to understand what kind of value are you emitting here, e.g

`
new EventEmitter<{ expressionValue: OperationProperty.ExpressionValue }>
`

client-ui/src/app/components/header/header.component.html
oleksandr-korchak: If it bothering you, just add some semi-transparent background, and it will be good enough for now for these edge cases.

client-ui/src/app/components/header/header.component.ts
oleksandr-korchak: What can I say ...

`
readonly varName: string = ''
`
is redundant in Typescript. 

So you can start thinking to not require it :)

I will mark that conversion as resolved.

client-ui/src/app/model/operation/operation-property.models.ts
oleksandr-korchak: I do not understand what we are talking about here. It does not break any principle as far we know what you are doing here. We are working with an object we have created by our own and which exists only in the current scope. It will not bring any unexpected behavior and any potential bugs, plus it solves a few problems: 
- it is better to read
- it is faster
- it uses less memory.

If you have a big problem with the mutating accumulator, then do not use reduce function here. As it is not a good use case in your case, then just work with forEach. But do not create a bunch of copy objects.

wajda: Let's discuss it on CQC

client-ui/src/app/components/lineage/lineage-details/operation-properties-details/extra-properties/extra-properties-expression.component.ts
wajda: it's still `...value.value`. But OK, let's do it.

client-ui/src/app/components/header/header.component.html
wajda: Not everybody has full HD :)

oleksandr-korchak: yes, you are right most of the people has higher resolution :)

oleksandr-korchak: I have nothing to say more about that. I will mark that conversion as resolved.

client-ui/src/app/model/operation/operation-property.models.ts
oleksandr-korchak: your reducer did not become significantly smaller, it is huge now. But you are iterating 2 times more now. 

I have nothing to say more here. I will mark that discussion as resolved.

client-ui/src/app/components/modal/modal-expression/modal-expression.component.less
oleksandr-korchak: Oh, you are calculating a class name in TS and call it className, did not mention that before.

It is not a very good idea to do so.
But I do not know if I should comment it in that case as you kind of add some piece of mess to another pice of a mess, where most staff are of type any. It is kind of quick fix and, I guess, it is good enough for the current state.

will mark that conversation as resolved.

client-ui/src/app/components/header/header.component.ts
wajda: Completely agree, for literals! But `__APP_VERSION__` isn't a literal, it is a global constant which type isn't obvious in the place of `appVersion`.

Issue #732, opened 2020-06-16
Visible URL to detailed lineage


DaimonPl:
It would be good if UI would additionally display URL to detailed lineage of job.

It took me really a while to realize that double click on job on graph has some action :D

wajda:
:facepalm: man, where have you been yesterday? :laughing: Now I need to release a new version of the entire project just to add
```<a href="..." >Show detailed lineage</a>```

DaimonPl:
I think it can safely wait for other changes :)

wajda:
+ fixed breadcrumbs navigation

PR #733, opened 2020-06-16
spline #732 Visible URL to detailed lineage

wajda:


client-ui/src/app/components/lineage-overview/lineage-overview-details/lineage-overview-details.component.html
```diff
@@ -36,10 +36,23 @@ <h3 class="text-center pb-3 mt-3"> Execution Event </h3>
 
     <div class="divider"></div>
 
+    <!-- EXECUTION PLAN DETAILS -->
+    <div *ngIf="executedLogicalPlan$ | async as executedLogicalPlan" class="exec-plan-details">
+      <div ngbDropdown class="d-inline-block toolbar">
```
cerveada, line 41: It doesn't feel right to have a dropdown for one item, but the whole button might not fit into the same line. Maybe new line with button would be better?

wajda: I tried... it sucks one way or another :( With a dropdown at least it's not that bulky, plus we have a decent place to add more actions if we need. 
I'll add a hint on the button to tell the user that the same action can be achieved by a double click on a node. So there will be less chance that a user would want to use it as a primary navigation control.

cerveada: OK, lets do it this way for now, but I think it would be good to revisit this later

wajda: We're gonna have a new UI with (hopefully) better navigation :)

Issue #741, opened 2020-06-25
Spline dashboard takes too long to load


wajda:
the query used to populate the dashboard table has O(n) or worse complexity. Should be O(1)

PR #750, opened 2020-07-10
spline #741 improve dashboard data loading performance

cerveada:
- performance improvements
- removing deprecated indices

### DIFF
--- a/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionEventRepositoryImpl.scala
+++ b/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionEventRepositoryImpl.scala
@@ -57,6 +57,8 @@ class ExecutionEventRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
         "asAtTime" -> (asAtTime: java.lang.Long)
       ))
 
+    val sortFieldMap = Map("executionEventId" -> "_key", "applicationId" -> List("extra", "appId"))
+
     val eventualArangoCursorAsync = db.queryAs[ExecutionEventInfo](
       """
         |FOR ee IN progress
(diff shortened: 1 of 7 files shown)


producer-services/src/main/scala/za/co/absa/spline/producer/service/repo/ExecutionProducerRepositoryImpl.scala
```diff
@@ -90,27 +90,59 @@ class ExecutionProducerRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) exte
       Map("keys" -> events.map(_.planId))
     )
 
-    val progressNodes = events.map(e => Progress(
+    val queryPerformanceData = db.queryAs[Map[String, Any]](
```
wajda, line 93: When naming a `Future` variables please use one of the following pattern (depending on what reads better in the context):
- `xxxFuture`
- `eventualXxx`
Same as you would do for `Option` identifiers, e.g. `maybeXxx` or `xxxOption`

wajda: Also the phrase _performance data_ reads to me as the data related to the performance, i.e. some stats, metrics etc.
I would rather call it `execPlanDetails` or alike.

persistence/src/main/scala/za/co/absa/spline/persistence/model/entities.scala
```diff
@@ -63,5 +63,12 @@ case class Progress(
   timestamp: Long,
   error: Option[Any],
   extra: Map[String, Any],
-  override val _key: String
+  override val _key: String,
+  // the rest of the parameters is copied from other entities - for performance reasons
```
wajda, line 67: I wonder if it wouldn't make sense to nest these fields into another one called say `execPlanDetails`? This would make the above comment unnecessary, and it would also look more clear from the places of usage.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionEventRepositoryImpl.scala
```diff
@@ -57,52 +57,48 @@ class ExecutionEventRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
         "asAtTime" -> (asAtTime: java.lang.Long)
       ))
 
+    val sortFieldMap = Map("executionEventId" -> "_key", "applicationId" -> List("extra", "appId"))
```
wajda, line 60: What about making it a constant in a companion object?

Issue #772, opened 2020-09-17
Admin CLI: Foxx services aren't installed when executing a JAR-file


wajda:
`java -jar admin-cli.jar db-upgrade ...` performs the migration as expected but the Foxx services aren't installed. No errors.

Related to #763

PR #774, opened 2020-09-17
spline #772 Admin CLI: Foxx services aren't installed when executing …

wajda:
…a JAR-file


commons/src/main/scala/za/co/absa/spline/common/rest/RESTClient.scala
```diff
@@ -0,0 +1,85 @@
...
+            }
+          })
+        (response.getStatusLine, maybeBody.orNull)
+      }
+
+    respStatusLine.getStatusCode match {
+      case 200 | 201 | 204 =>
+        Future.successful(respBody)
+      case _ =>
+        throw new HttpStatusException(respStatusLine.getStatusCode, s"ArangoDB response: $respStatusLine. $respBody")
```
cerveada, line 82: Why not use Future.failed() instead of throwing?

Or opposite question. Is the Future actually useful here? We could return `respBody` immediately.

wajda: Good point. Actually I'd prefer to keep async nature of the methods. Even though the current implementation uses a blocking Http client, wrapping it with a future makes it easy to change the impl anytime.

Would something like this read better?
```scala
    Future.fromTry(Try(respStatusLine.getStatusCode match {
      case 200 | 201 | 204 =>
        respBody
      case _ =>
        throw new HttpStatusException(respStatusLine.getStatusCode, s"ArangoDB response: $respStatusLine. $respBody")
    }))
```

wajda: `Future.fromTry(Try` can even be placed on the beginning of the method

cerveada: This would be the easiest to read for me:
```
    respStatusLine.getStatusCode match {
      case 200 | 201 | 204 =>
        Future.successful(respBody)
      case statusCode =>
        Future.failed(new HttpStatusException(statusCode, s"ArangoDB response: $respStatusLine. $respBody"))
    }
```

But wrapping the whole method in `Future.fromTry(Try` would catch all exceptions happening there, but I don't know whether that is something we want.

Both options are ok for me.

wajda: I'd say if a method returns a `Future` then it shouldn't throw. It often leads to inconsistency and difficulties in error handling.

wajda: done

commons/src/main/scala/za/co/absa/spline/common/rest/RESTClient.scala
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.common.rest
+
+import scala.concurrent.Future
+
+trait RESTClient {
+  @throws[HttpStatusException]
```
cerveada, line 22: This might be a bit confusing now

cerveada (dismissed):

commons/src/main/scala/za/co/absa/spline/common/rest/RESTClient.scala
wajda: removed

cerveada (approved):

Issue #793, opened 2020-10-20
Introduce an AffectsDataSrc And DependsDataSrc  Endpoint


mohanajuhi166:
## Background
We have the existing collections - dataSource and executionPlan in **arangodb** collections. There is a small opportunity to introduce a minor feature, which can provide us the information on - which dataSources are affected and on which datasources its is dependent. 

## Feature & Proposed Solution
We have two ways of doing it - 

**Option 1**
We can create a new graph called affectsGraph  - with the following
@From - executionPlan
@To - DataSource

And then using the above graph for query. 

Example : 
```aql
FOR v,e,p IN INBOUND 'dataSource/<id>' GRAPH 'affectsGraph'
    RETURN DISTINCT v.extra.appName
```

**Option 2**
Introduce a new controller and new repository implemention - called affectsRepositoryImpl.scala 
using the below query
```aql
FOR e IN executionPlan, a IN affects OUTBOUND @dataSrc GRAPH 'affectsGraph'
    PRUNE a._to == @dataSrc AND e._id == a._from
    RETURN DISTINCT e.extra.appName
```
## Example 
<img width="873" alt="git_spline" src="https://user-images.githubusercontent.com/19162354/96572193-2c4fb300-12ea-11eb-8913-9d03f2400fd0.png">

Referring to the image above as example

The dependsDataSrc endpoint will return **WF_READ_1 & WF_READ_2**
The affectsDataSrc endpoint will return **WF_WRITE_1**
For the same,I would like to propose a minor feature addition for which I have coded the solution and would like to raise a pull request.

Kindly let me know your thoughts on the same.

wajda:
Hi @mohanajuhi166,
Let me understand your proposal. You said the goal is to track the dependencies between execution plans and data source, and for that you propose to introduce another graph. How are your edges would differ from existing `depends` and `affects` edges that have the very same purpose?

mohanajuhi166:
Hi @wajda , thanks a lot for replying. I tried the option 1 (creating a graph) to use pruning condition and potentially a run time complexity reduction but I am not very sure of the same. 

I would like to propose the option 2 - which is introducing two endpoints. to provide the information on dependant and affected workflow. This includes a repository implementation with calling query inside **consumer-rest-core**

Let me know your thoughts on the same.

wajda:
That makes sense.
Let's use the following resource URLs:
[...]

PR #798, opened 2020-11-05
Feature Request : Introduction of An DependsDataSrc Endpoint

mohanajuhi166:
Feature Request Part one for the following

/data-sources/{ds_id}/execution-plan-ids -- a collection of the execution plan IDs directly associated with the given data source.

Feature Issue : https://github.com/AbsaOSS/spline/issues/793

### DIFF
--- a/producer-services/src/main/scala/za/co/absa/spline/producer/service/repo/ExecutionProducerRepository.scala
+++ b/producer-services/src/main/scala/za/co/absa/spline/producer/service/repo/ExecutionProducerRepository.scala
@@ -24,5 +24,4 @@ trait ExecutionProducerRepository {
   def insertExecutionPlan(executionPlan: ExecutionPlan)(implicit ec: ExecutionContext): Future[Unit]
   def insertExecutionEvents(executionEvents: Array[ExecutionEvent])(implicit ec: ExecutionContext): Future[Unit]
   def isDatabaseOk()(implicit ec: ExecutionContext): Future[Boolean]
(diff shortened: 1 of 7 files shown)


producer-services/src/main/scala/za/co/absa/spline/producer/service/repo/ExecutionProducerRepository.scala
```diff
@@ -24,4 +24,5 @@ trait ExecutionProducerRepository {
   def insertExecutionPlan(executionPlan: ExecutionPlan)(implicit ec: ExecutionContext): Future[Unit]
   def insertExecutionEvents(executionEvents: Array[ExecutionEvent])(implicit ec: ExecutionContext): Future[Unit]
   def isDatabaseOk()(implicit ec: ExecutionContext): Future[Boolean]
+  def getDataSources(value: String, access: Optional[String])(implicit ec: ExecutionContext): Future[Array[String]]
```
cerveada:
```suggestion
  def getDataSources(value: String, access: Option[String])(implicit ec: ExecutionContext): Future[Array[String]]
```
- Optional is Java class
- Option is Scala class

wajda:
I think it should be a part of Consumer API, not Producer API. Could you please move it?

producer-services/src/main/scala/za/co/absa/spline/producer/service/repo/ExecutionProducerRepository.scala
mohanajuhi166: making the change

mohanajuhi166:
Adding the changes as per requested

PR #799, opened 2020-11-07
Feature Request : Introduction of An DependsDataSrc Endpoint

mohanajuhi166:
Feature Request Part one for the following

/data-sources/{ds_id}/execution-plan-ids -- a collection of the execution plan IDs directly associated with the given data source.

Feature Issue : #793
Note : Created new PR due to some issues  in my fork

Added option for scala
Moved code to consumer

### DIFF
--- a/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
+++ b/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
@@ -66,7 +66,7 @@ class LineageDetailedController @Autowired()(
       AttributeLineageAndImpact(maybeAttrLineage, attrImpact)
     })
 
-  @GetMapping(value = Array("/execution-plans/{plan_id}/data-sources"))
+  @GetMapping(value = Array("execution-plans/{plan_id}/data-sources"))
   @ResponseStatus(HttpStatus.OK)
   def getAffectsByDataSrc(@PathVariable("plan_id") planId: String
     @ApiParam(value= "access")
--- a/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
+++ b/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
@@ -20,12 +20,13 @@ import java.util.UUID
 
 import io.swagger.annotations._
 import org.springframework.beans.factory.annotation.Autowired
-import org.springframework.web.bind.annotation._
 import org.springframework.http.HttpStatus
+import org.springframework.web.bind.annotation._
 import za.co.absa.spline.consumer.rest.controller.LineageDetailedController.AttributeLineageAndImpact
 import za.co.absa.spline.consumer.service.attrresolver.AttributeDependencyResolver
 import za.co.absa.spline.consumer.service.internal.AttributeDependencySolver
-import za.co.absa.spline.consumer.service.model.{AttributeGraph, ExecutionPlanInfo, LineageDetailed}
+import za.co.absa.spline.consumer.service.model.AccessValue.AccessValue
+import za.co.absa.spline.consumer.service.model.{AccessValue, AttributeGraph, ExecutionPlanInfo, LineageDetailed}
 import za.co.absa.spline.consumer.service.repo.ExecutionPlanRepository
 
 import scala.concurrent.Future
@@ -67,14 +68,20 @@ class LineageDetailedController @Autowired()(
       AttributeLineageAndImpact(maybeAttrLineage, attrImpact)
     })
 
+  def loadAccess(access: String): Option[AccessValue] = {
+    AccessValue.values.find(_.toString == access)
+  }
+
   @GetMapping(value = Array("execution-plans/{plan_id}/data-sources"))
   @ResponseStatus(HttpStatus.OK)
-  def getAffectsByDataSrc(@PathVariable("plan_id") planId: String,
-    @ApiParam(value= "access")
-    @RequestParam("access")access: Option[String]
+  def execPlanDataSources(
+    @PathVariable("plan_id") planId: String,
+    @ApiParam(value = "access")
+    @RequestParam(name = "access", required = false) access: String
   ): Future[Array[String]] = {
-    import scala.concurrent.ExecutionContext.Implicits._
-    repo.getDataSources(planId, access)
+
+    val param = loadAccess(access)
+    repo.getDataSources(planId, param)
   }
 }
 
--- a/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
+++ b/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
@@ -21,6 +21,7 @@ import java.util.UUID
 import io.swagger.annotations._
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.web.bind.annotation._
+import org.springframework.http.HttpStatus
 import za.co.absa.spline.consumer.rest.controller.LineageDetailedController.AttributeLineageAndImpact
 import za.co.absa.spline.consumer.service.attrresolver.AttributeDependencyResolver
 import za.co.absa.spline.consumer.service.internal.AttributeDependencySolver
--- a/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
+++ b/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
@@ -25,8 +25,7 @@ import org.springframework.web.bind.annotation._
 import za.co.absa.spline.consumer.rest.controller.LineageDetailedController.AttributeLineageAndImpact
 import za.co.absa.spline.consumer.service.attrresolver.AttributeDependencyResolver
 import za.co.absa.spline.consumer.service.internal.AttributeDependencySolver
-import za.co.absa.spline.consumer.service.model.AccessValue.AccessValue
-import za.co.absa.spline.consumer.service.model.{AccessValue, AttributeGraph, ExecutionPlanInfo, LineageDetailed}
+import za.co.absa.spline.consumer.service.model.{AttributeGraph, DataSourceActionType, ExecutionPlanInfo, LineageDetailed}
 import za.co.absa.spline.consumer.service.repo.ExecutionPlanRepository
 
 import scala.concurrent.Future
@@ -68,8 +67,8 @@ class LineageDetailedController @Autowired()(
       AttributeLineageAndImpact(maybeAttrLineage, attrImpact)
     })
 
-  def loadAccess(access: String): Option[AccessValue] = {
-    AccessValue.values.find(_.toString == access)
+  def loadAccess(access: String): Option[DataSourceActionType] = {
+    DataSourceActionType.values.find(_.toString == access)
   }
 
   @GetMapping(value = Array("execution-plans/{plan_id}/data-sources"))
@@ -80,8 +79,8 @@ class LineageDetailedController @Autowired()(
     @RequestParam(name = "access", required = false) access: String
   ): Future[Array[String]] = {
 
-    val param = loadAccess(access)
-    repo.getDataSources(planId, param)
+    val dataSourceActionTypeOption = loadAccess(access)
+    repo.getDataSources(planId, dataSourceActionTypeOption)
   }
 }
 
--- a/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
+++ b/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
@@ -65,6 +65,16 @@ class LineageDetailedController @Autowired()(
 
       AttributeLineageAndImpact(maybeAttrLineage, attrImpact)
     })
+
+  @GetMapping(value = Array("/execution-plans/{plan_id}/data-sources"))
+  @ResponseStatus(HttpStatus.OK)
+  def getAffectsByDataSrc(@PathVariable("plan_id") planId: String
+    @ApiParam(value= "access")
+    @RequestParam("access")access: Option[String]
+  ): Future[Array[String]] = {
+    import scala.concurrent.ExecutionContext.Implicits._
+    repo.getDataSources(planId, access)
+  }
 }
 
 object LineageDetailedController {
--- a/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
+++ b/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
@@ -68,7 +68,7 @@ class LineageDetailedController @Autowired()(
 
   @GetMapping(value = Array("execution-plans/{plan_id}/data-sources"))
   @ResponseStatus(HttpStatus.OK)
-  def getAffectsByDataSrc(@PathVariable("plan_id") planId: String
+  def getAffectsByDataSrc(@PathVariable("plan_id") planId: String,
     @ApiParam(value= "access")
     @RequestParam("access")access: Option[String]
   ): Future[Array[String]] = {
--- a/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/AccessValue.scala
+++ b/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/AccessValue.scala
@@ -1,0 +1,9 @@
+package za.co.absa.spline.consumer.service.model
+
+object AccessValue extends Enumeration {
+  type AccessValue = Value
+
+  val read = Value("read")
+  val write = Value("write")
+  val default = Value("default")
+}
--- a/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/AccessValue.scala
+++ b/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/AccessValue.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2020 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package za.co.absa.spline.consumer.service.model
 
 object AccessValue extends Enumeration {
--- a/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/AccessValue.scala
+++ b/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/AccessValue.scala
@@ -5,5 +5,4 @@ object AccessValue extends Enumeration {
 
   val read = Value("read")
   val write = Value("write")
-  val default = Value("default")
 }
--- a/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/DataSourceActionType.scala
+++ b/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/DataSourceActionType.scala
@@ -16,9 +16,13 @@
 
 package za.co.absa.spline.consumer.service.model
 
-object AccessValue extends Enumeration {
-  type AccessValue = Value
+sealed trait DataSourceActionType
 
-  val read = Value("read")
-  val write = Value("write")
+object DataSourceActionType {
+
+  case object Read extends DataSourceActionType
+
+  case object Write extends DataSourceActionType
+
+  val values = Seq(Read, Write)
 }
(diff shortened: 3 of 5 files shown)


consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
```diff
@@ -65,6 +66,16 @@ class LineageDetailedController @Autowired()(
...
       AttributeLineageAndImpact(maybeAttrLineage, attrImpact)
     })
+
+  @GetMapping(value = Array("execution-plans/{plan_id}/data-sources"))
+  @ResponseStatus(HttpStatus.OK)
+  def getAffectsByDataSrc(@PathVariable("plan_id") planId: String,
+    @ApiParam(value= "access")
+    @RequestParam("access")access: Option[String]
+  ): Future[Array[String]] = {
+    import scala.concurrent.ExecutionContext.Implicits._
```
wajda: The global execution context is already in the scope of visibility, so this import is redundant.

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
```diff
@@ -65,6 +66,16 @@ class LineageDetailedController @Autowired()(
 
       AttributeLineageAndImpact(maybeAttrLineage, attrImpact)
     })
+
+  @GetMapping(value = Array("execution-plans/{plan_id}/data-sources"))
+  @ResponseStatus(HttpStatus.OK)
+  def getAffectsByDataSrc(@PathVariable("plan_id") planId: String,
```
wajda: nit: code styling:
as the common Scala code style guide suggests if method parameters can't be put all on one line, they should be put one each per line, starting on the next line.
```scala
def getAffectsByDataSrc(
  @PathVariable("plan_id") planId: String,
  ...
```

wajda: Can you find a better name for the method? To me `getAffectsByDataSrc` doesn't seem to correlate with the resource URI `execution-plans/{plan_id}/data-sources`.

I'd suggest `execPlanDataSources`, or even just `dataSources`

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
```diff
@@ -65,6 +66,16 @@ class LineageDetailedController @Autowired()(
...
+    @ApiParam(value= "access")
+    @RequestParam("access")access: Option[String]
```
wajda: nit: missing space characters

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -134,6 +134,41 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
           edges = edges)
     }
   }
+
+  override def getDataSources(value: String, access: Option[String])(implicit ec: ExecutionContext): Future[Array[String]] ={
```
wajda: nit: missing space character

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -134,6 +134,41 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
...
+    )
+
+    val totalResult = readResult.zip(writeResult).map{case(listOne,listTwo) => listOne ++ listTwo}
+
+    if(access.isEmpty)
+      totalResult.map(_.toArray)
+    else if(access.equals("read"))
+      readResult.map(_.toArray)
+    else
+      writeResult.map(_.toArray)
```
wajda: nit: missing space characters: `if(`, `listOne,listTwo`, `map{case`

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -134,6 +134,41 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
           edges = edges)
     }
   }
+
+  override def getDataSources(value: String, access: Option[String])(implicit ec: ExecutionContext): Future[Array[String]] ={
+    val readResult = db.queryStream[String](
+      """
+        FOR d in dataSource
```
wajda: missing `|`

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -134,6 +134,41 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
...
+        |FOR dp in depends
+        |FILTER dp._from == @planId AND dp._to == d._id
+        |RETURN d.uri
+        |""".stripMargin
```
wajda: Duplicated `stripMargin`

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -134,6 +134,41 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
...
+        .stripMargin,
+      Map("planId" -> "executionPlan".concat("/").concat(value)
```
wajda: We prefer string interpolation to string concatenation.
```suggestion
      Map("planId" -> s"executionPlan/$value")
```

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -134,6 +134,41 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
...
+
+    val writeResult = db.queryStream[String](
+      """
+        FOR d in dataSource
+        |FOR aff in affects
+        |FILTER aff._from == @planId AND aff._to == d._id
+        |RETURN d.uri
+        |""".stripMargin
+        .stripMargin,
+      Map("planId" -> "executionPlan".concat("/").concat(value)
```
wajda: This is not the best way for querying a graph in ArangoDB.
Your query involves two scans with filter (full collection scan + index scan + filter).
In this case a graph traversal would be way more efficient.
Also it's advisable to use document keys (without collection name prefix) instead of full IDs outside the database. In this query you know that you are looking at the `executionPlan` collection, so passing it as a parameter makes no sense.

Consider the following query:
```aql
FOR ds IN 1..1 
    OUTBOUND DOCUMENT('executionPlan', @planKey) affects
    RETURN ds.uri
```

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -134,6 +134,41 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
...
+      )
+    )
+
+    val totalResult = readResult.zip(writeResult).map{case(listOne,listTwo) => listOne ++ listTwo}
+
+    if(access.isEmpty)
+      totalResult.map(_.toArray)
+    else if(access.equals("read"))
```
wajda: Consider using enums for this

wajda (changes requested):

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
mohanajuhi166: noted, making change

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
mohanajuhi166: done

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
mohanajuhi166: added

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: added

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: added

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: added

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: removed

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: added

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: makes sense. I am making the change.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: added , I added a separate enumerator class inside persistance/model . Let me know if I need to add it somewhere else. Thanks

mohanajuhi166:
@wajda I have added the requested changes. Let me know if any additional changes are required.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -135,36 +136,30 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
...
-        |RETURN d.uri
-        |""".stripMargin
-        .stripMargin,
-      Map("planId" -> "executionPlan".concat("/").concat(value)
-      )
+        |FOR ds IN 1..1
+        |OUTBOUND DOCUMENT('executionPlan', @planId) depends
+        |RETURN ds.uri
+        |""".stripMargin,
+      Map("planId" -> s"executionPlan/$value")
```
wajda: `executionPlan` collection is already specified in the query, so no need for the prefix here.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -135,36 +136,30 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
     }
   }
 
-  override def getDataSources(value: String, access: Option[String])(implicit ec: ExecutionContext): Future[Array[String]] ={
+  override def getDataSources(value: String, access: Option[AccessValue])(implicit ec: ExecutionContext): Future[Array[String]] = {
```
wajda: Could you please find a more descriptive name for the `value` parameter?

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -135,36 +136,30 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
...
+    val totalResult = readResult.zip(writeResult).map{ case(listOne,listTwo) => listOne ++ listTwo }
 
-    if(access.isEmpty)
+    if (access.isEmpty)
       totalResult.map(_.toArray)
-    else if(access.equals("read"))
+    else if (access.get.equals(AccessValue.read))
       readResult.map(_.toArray)
     else
       writeResult.map(_.toArray)
```
wajda: Sorry I didn't make it clear. One of the biggest benefits of using enums is that you can apply exhaustive pattern-matching instead of `if-else` blocks. Also it's a known best practice to use sealed traits with case objects instead of native Scala enums, primarily to be able to use _exhaustive_ pattern-matching. See for instance [this blog post](https://medium.com/@yuriigorbylov/scala-enumerations-hell-5bdba2c1216). Or take a look at `OnDBExistsAction` class in our project.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/AccessValue.scala
```diff
@@ -0,0 +1,9 @@
+package za.co.absa.spline.consumer.service.model
+
+object AccessValue extends Enumeration {
+  type AccessValue = Value
+
+  val read = Value("read")
```
wajda: Please use capitalized names for constant values (`Read` instead of `read`) according to [Scala naming convention](https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods)

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
wajda: Another observation - you execute both _depends_ and _affects_ queries on every method call regardless of what was actually requested. It has better be done lazily, so only necessary queries are executed.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/AccessValue.scala
wajda: Also please use more descriptive name for enumeration type. E.g. `AccessType`? Or maybe `DataSourceActionType`?

mohanajuhi166: Added

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: @wajda what should one return when no access type is provided? I was providing both affects and depends by default and hence calculating it again
do you mean, we should use lazy val? 
Thanks

mohanajuhi166: sharing a code snippet for search
`def getDataSourceResultFromActionType(
    actionType: DataSourceActionType,
    readResult: Future[Stream[String]],
    writeResult : Future[Stream[String]]): Future[Stream[String]] = actionType match {
    case Read => readResult
    case Write => writeResult
  }`

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: done

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
mohanajuhi166: Following your suggestions, this is what i came up with .
If accessType is empty, we return overview results.
What do you think of the the below implementation 
thanks

```
if(access.isEmpty)
      db.queryStream[String](
        """
          |FOR ds IN 1..1 OUTBOUND DOCUMENT ('executionPlan', '20e4a7d1-610b-459a-97b9-cff7aac4b1ab')  GRAPH 'overview'
          | RETURN ds.uri
          |""".stripMargin,
        Map("planId" -> s"$execPlanId")
      ).map(_.toArray)`




else
    access.get match {
      case Read =>  db.queryStream[String](
        """
          |FOR ds IN 1..1
          |OUTBOUND DOCUMENT('executionPlan', @planId) depends
          |RETURN ds.uri
          |""".stripMargin,
        Map("planId" -> s"$execPlanId")
      ).map(_.toArray)

    

      case Write => db.queryStream[String](
        """
          |FOR ds IN 1..1
          |OUTBOUND DOCUMENT('executionPlan', @planId) affects
          |RETURN ds.uri
          |""".stripMargin,
        Map("planId" -> s"$execPlanId")
      ).map(_.toArray)
    }
  }
```

wajda: `lazy val` could be a solution, yes. Your last snippet would work as well. Yet another solution would be a dynamically generated query. Whatever you prefer.

Regarding handling empty `accessType` I think that's OK. Although personally I'd rather pass `Seq[AccessType]` instead of `Option[AccessType]` and would return empty result if no access type is provided. But that's might be a matter of taste, so feel free to decide.

The code style-wise there are two things that caught my attention in your last code piece:
1. `s"$execPlanId"` is just `execPlanId`
2. `access.get` / `access.isEmpty` - using the combination of those method is considered a bad practice. Try using more idiomatic approach to _Option_ handling. E.g.
```scala
access
  .map({
    case Read => ...
    case Write => ...
  })
  .getOrElse({
    ...
  })
```

wajda: And one more little thing - in your AQL using `depends, affects` would be more readable and understandable than `GRAPH 'overview'` IMO. Although technically it's the same thing, in the current context where you have three similar queries that only differ in selected edges, it just feels more natural to stick with edge names instead of graph names.

mohanajuhi166: I added my last snippet
You are absolutely right about the graph overview , didnt strike me that time. Thanks for the tip.

I kept it as option and i am returning both results if no access type is provided instead of an empty list. Hope that works ?
Changed the variable names to proper names.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/model/DataSourceActionType.scala
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.consumer.service.model
+
+sealed trait DataSourceActionType
+
+object DataSourceActionType {
```
mohanajuhi166, line 21: changed name as per recommendation

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepository.scala
```diff
@@ -18,12 +18,14 @@ package za.co.absa.spline.consumer.service.repo
...
+import za.co.absa.spline.consumer.service.model.{DataSourceActionType, LineageDetailed}
 
 import scala.concurrent.{ExecutionContext, Future}
 
 trait ExecutionPlanRepository {
   def findById(execId: Id)(implicit ec: ExecutionContext): Future[LineageDetailed]
 
   def loadExecutionPlanAsDAG(execId: Id)(implicit ec: ExecutionContext): Future[ExecutionPlanDAG]
+
+  def getDataSources(execPlanId: String, access: Option[DataSourceActionType])(implicit ec: ExecutionContext): Future[Array[String]]
```
mohanajuhi166, line 30: changed name as per recommendation

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -134,6 +135,39 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
...
+            |RETURN ds.uri
+            |""".stripMargin,
+          Map("planId" -> execPlanId)
+        ).map(_.toArray)
+      })
+      .getOrElse({
+        db.queryStream[String](
+          """
+            |FOR ds IN 1..1
+            |OUTBOUND DOCUMENT('executionPlan', @planId) affects, depends
```
mohanajuhi166, line 164: removed  graph overview

wajda (approved):
LGTM

wajda:
Rebasing it and closing. New PR is #802

wajda:
Merged to develop. Thank you for your contribution!

mohanajuhi166:
@wajda thanks for all your help
I will try to contribute more in the near future.

PR #803, opened 2020-11-24
Feature Request : Introduction of An execution Plan Name Endpoint : 2nd Part

mohanajuhi166:
From feature request : https://github.com/AbsaOSS/spline/issues/793

This is the second API : `/data-source/{ds_id}/execution-plan-ids?access=write` -- returns an array of the exec plans that writes into the given source.

I have changed the API signature to return names of execution plan instead of ids

`/data-source/{ds_id}/execution-plan-name`

- Read 
- Access
- If none provided, supplies information on both depends and affects


wajda:
I think we need to restructure the URI to comply with RESTful API design best practices.

A URI should represent a resource. `/data-source/123/execution-plan-names` (or `-ids`) is nether a resource nor a property of _Data Source_ entity. It rather is a projection of _Execution Plan_ top-level entity collection - selected by a matching criteria.
So I'd rather query it using a DSL like this:
```http
GET /execution-plans?affect-ds-id=123&fields=id,name
```

mohanajuhi166:
@wajda , to have both depends and affects
can we have something like below :-

`GET /execution-plans?ds-id=123&access=affects&field=name or id `

than in implementation , we can have the below : -
```
val fieldVal = (field: @switch) match {
      case name => "execPlan.extra.appName"
      case id => "execPlan._id"
      case _  => "execPlan.extra"
    }
```

and use it in below query 
```
"""
            |FOR execPlan IN 1..1
            |INBOUND DOCUMENT('dataSource', @dsId) depends
            |RETURN @fieldId
            |""".stripMargin,
          Map("dsId" -> dsId, "fieldId" -> fieldVal)
        )
```

what do you think of the same?

mohanajuhi166:
@wajda , to have both depends and affects  can we have something like below :-
`GET /execution-plans?ds_id=123&access=Read&field=name or id`
than in implementation , we can have the below : -
```

val fieldVal = field match {
      case name => "execPlan.extra.appName"
      case id => "execPlan._id"
      case _  => "execPlan.extra"
    }
```
and use it in below query

```
case Read =>  db.queryStream[String](
          s"""
            |FOR execPlan IN 1..1
            |INBOUND DOCUMENT('dataSource', @dsId) depends
            |RETURN $fieldVal
            |""".stripMargin,
          Map("dsId" -> dsId)
        ).map(_.toArray)
```
what do you think of the same?

updated the PR accordingly

wajda:
As a _data source ID_ and an _access_ properties work in pair I'd rather keep them coupled. Also it would be beneficial to allow for specifying multiple data source relations (e.g. find plans that read from A and write to B). 
What if we do this?
```http
GET /execution-plans/?ds-rel=read:111,read:222,write:333
```
Also could you please allow for multiple `field` values. Either make it an array of string, or make it plural accepting comma-separated values.

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanRepositoryImpl.scala
```diff
@@ -169,33 +169,43 @@ class ExecutionPlanRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends
...
-  override def getExecutionPlan(dsId: String, access: Option[DataSourceActionType])(implicit ec: ExecutionContext): Future[Array[String]] =  {
+  override def getExecutionPlan(
+   dsId: String,
+   access: Option[DataSourceActionType],
+   field: String)
+  (implicit ec: ExecutionContext): Future[Array[String]] =  {
+    val fieldVal = field match {
+      case name => "execPlan.extra.appName"
+      case id => "execPlan._id"
+      case _  => "execPlan.extra"
```
wajda: Did you mean?
```suggestion
      case "name" => "execPlan.extra.appName"
      case "id" => "execPlan._id"
      case _  => "execPlan.extra"
```

wajda: Also consider keeping a query variable `execPlan` inside the query. E.g. `RETURN execPlan.$path`

mohanajuhi166: sure, will do.

pratapmmmec:
@wajda 
Just want to know with "GET /execution-plans/?ds-rel=read:111,read:222,write:333", what use case are we trying to solve and how will be represented on UI? 
We also have a use case where we want to do impact analysis for any table. For instance, if we need to add, drop, alter a column in any table the which all workflows (reading from and writing to it) might be impacted. It will be good we can have an endpoint which can solve this use case. Next step/feature can be to highlight only the impacted workflow on selecting a particular column for that table.

Any thoughts on this?

wajda:
> what use case are we trying to solve

I thought you wanted to get exec plans by their relation to data sources. To me it's a generic functionality and makes it a good addition to Spline Consumer API. Your use-case is fine, but we need to think broader to accommodate similar use-cases the other users might have. That's why I'm asking to at least allow for multiple values in your filter, not just one.

pratapmmmec:
> it's a generic functionality and makes it a good addition to Spline Consumer API
>
Probably I couldn't put my question correctly. As already pointed by you and it makes perfect sense to look for more generic use case instead of solving a specific one. Can you please confirm below:
1. Are these parameters evaluated as OR or AND? 
2. For specific use case of finding read and write dependent plans, do we then need to invoke same endpoint with two different params like below

-       GET /execution-plans/?ds-rel=read:111
-       GET /execution-plans/?ds-rel=write:111

mohanajuhi166:
@wajda I added the additional changes to the code . I am not sure if its the standard way to get the results. 
Will make the requested changes to change it.
Also, do we need to format the final result as. I have we are displaying multiple fields? What library should we use for json parsing, 

Thank you.

wajda:
> OR or AND

Good point. What do you think?
To me OR seems to be more intuitive for comma separated values. AND could be encoded by adding more `&ds-rel=` parameters. For NOT we can use `!` symbol before the value.

For instance `(read:A OR read:B) AND NOT write:C` could be represented as
```http
GET .../?ds-rel=read:A,read:B&ds-rel=!write:C
```

We could also implement some DSL with respective keywords, but it would probably be an overkill.

wajda:
As for your last commit I would say because of the query generation logic is getting more complex, it deserves a separate query builder class covered with a set of unit tests.

Regarding the code style - generally we prefer functional paradigm, immutable collections and _val_'s over _var_'s if possible. But if you feel more comfortable with mutable algorithms it's OK as long as it meets common code standards (SOLID) and is covered with tests.

A few findings:
1. Watch for code injection vulnerability risk when composing a query. All input data must be inserted to a query via query binding parameters (`@xyz`), never via string concatenation (`$xyz`)
1. Think about returning exec plan IDs from the sub-queries, not full plans. When you get the final exec plan ID list you can fetch the docs by doing another FOR. The reason is that it might be more efficient in case of overlapping set of exec plans returned by different sub-queries executed potentially in parallel. But it has better be verified in Arangodb query profiler.
1. `INTERSECTION()` function accepts minimum 2 arguments
1. User query validation. REST endpoint should return 400 on badly composed query. The `ds_rel` parsing has to be done on the controller level and the service method should accept typed input.

> What library should we use for json parsing

Not sure why you need a JSON parser. If you mean serializer then you don't need to use any.
The framework does it for you. Just return a case-class or a collection of case-classes, or a Map from a controller (or a Future of those), and it will be correctly serialized to JSON.

mohanajuhi166:
@wajda I am thinking to implement the OR functionality first (comma separated), explained below:-

We can have a separate function to query result for each option - ex read:1111. Result returned will be executionIds

Then I can run a separate query as below for the executionIds.

```aql
FOR execPlan IN executionPlan
FILTER execPlan._id IN @execPlanIds
RETURN { @fieldValuesBody }
```

I thought of this because, query building wont be that complex, and when we get the list of execPlanIds, we can run a separate query for the same.
getExecutionPlanIdForDataSourceRelation will give us executionPlanId for each pair - ex : read:111


```
for (dsRel <- datasourceRelation) {

          val dsRelPair = dsRel.split(":")
          if(!dsRelPair.isEmpty && dsRelPair.size == 2) {
            queryBuilderForExecutionPlan += getExecutionPlanIdForDataSourceRelation(dsRelPair);
          }
       }
    val executionPlanIds =   queryBuilderForExecutionPlan //get the ids

    for (field <- fields) {
      field match{
        case "name" => fieldValuesBuilder += "name: execPlan.extra.appName"
        case "id" => fieldValuesBuilder += "id : execPlan._id"
        case _  => fieldValuesBuilder += "extra : execPlan.extra"
      }
    }

    db.queryStream[String](
      """
        |FOR execPlan IN executionPlan
        |FILTER execPlan._id IN @execPlanIds
        |RETURN { @fieldValuesBody }
        |""".stripMargin,
      Map("executionPlanBody" -> executionPlanIds.toArray, "fieldValuesBody" -> fieldValuesBuilder.result().mkString(","))
    ).map(_.toArray)

  }
```

What do you think?

wajda:
I would prefer to execute one big query instead of N smaller ones, to give AQL optimizer better control on things.
If composing a single query is hard (which I don't think it actually is for boolean combinators) then Foxx service could be an alternative. The reason why I don't like multiple queries executed from the application layer is because of transferring data from a DB to app level and back. Even though you would expect a few objects in the final result, the intermediate sub-queries could return large amount of data in big databases.

mohanajuhi166:
Sure, will do that. For now, I will implement the below OR functionality . And then move to AND and NOT

for OR

combining results for OR like below
```
LET results1 = (FOR execPlan IN 1..1
                     INBOUND DOCUMENT('dataSource','id1') depends
                     RETURN execPlan._id)
LET results2 =(FOR execPlan IN 1..1
                     INBOUND DOCUMENT('dataSource','id2') affects
                     RETURN execPlan._id)
                     
FOR x IN UNION(results1, results2)
RETURN x
```

Thoughts?

wajda:
Sure. I'll review it when it's ready. Take your time. Thanks.

mohanajuhi166:
@wajda  thanks for the above. Can you point me to what kind of unit tests can I write for the function? 
I am not entirely sure on the same
Also, I updated the code with validation check at controller and function for one single query, let me know your thoughts on the same.
I have currently used errorControllerAdvice and throwing an exception. Does that work?
Do we need to use scalaz validation itself on request parameter ? Or constraint validation in java should be used? can I use 
http://wix.github.io/accord/
I am a bit new to Scala so trying to still optimise the code.
Thank you

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/ExecutionPlanDetailsRepositoryImpl.scala
```diff
@@ -0,0 +1,105 @@
...
+                   |INBOUND DOCUMENT('dataSource',@dsId$resultIndex) depends
+                   |RETURN execPlan)
+                   """.stripMargin
+
+              queryBuilderForExecutionPlan += readQuery
+
+            case "write" =>
+              val writeQuery =
+                s"""
+                   |LET result$resultIndex = (FOR execPlan IN 1..1
```
mohanajuhi166, line 71: I am not sure if there is another way to use @resultId , I tried doing that via bind parameters but it didnt work

I am creating various variables like for ex: dsId 0,dsId1 etc using @dsId$resultIndex

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
```diff
@@ -77,6 +79,33 @@ class LineageDetailedController @Autowired()(
...
+
+  @GetMapping(value = Array("data-sources"))
+  @ResponseStatus(HttpStatus.OK)
+  def dataSourcesExecPlan(
+    @ApiParam(value = "ds_rel")
+    @RequestParam(name = "ds_rel", required = true) datasourceRelation: Array[String],
+    @ApiParam(value = "fields")
+    @RequestParam(name = "fields", required = true) fields: Array[String]
+  ): Future[Array[Map[String, Any]]] = {
+    if(!datasourceRelListIsValid(datasourceRelation))
```
mohanajuhi166, line 104: ignore on this part, I am working on it

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageDetailedController.scala
```diff
@@ -77,6 +78,20 @@ class LineageDetailedController @Autowired()(
     val dataSourceActionTypeOption = DataSourceActionType.findValueOf(access)
     repo.getDataSources(planId, dataSourceActionTypeOption)
   }
+
+  @GetMapping(value = Array("data-sources"))
```
wajda, line 82:
```suggestion
  @GetMapping(value = Array("execution-plans"))
```
We return execution plans details.

wajda:
Sorry @mohanajuhi166, we have to put this on-hold for a month or two, due to lack of capacity.

mohanajuhi166:
@wajda , no issues at all
Can you let me know about the tests I need to provide so that I can work on it in the meantime ?
Also, not exactly sure , what is the capacity in this context ?
Thank you

wajda:
I mean the dev team capacity. All staff is 100% busy on more priority things. So there will be most likely nobody to assist you or review your PR until about February.
Sure, you can continue working on it. We'll definitely return to it later, as more work on Consumer API has to come anyway.
Regarding testing, we have no coverage requirements. The tests just has to cover a decent portion of the most complex and error prone code, which is usually the one that involve some logical. A query builder is one of those and needs to be covered with unit-tests as much as possible.
Also an integration test against a real ArangoDB would be awesome. We can add additional steps to our CI pipeline to spin up ArangoDB in a docker before the tests. So you can assume there will be arangodb running on "localhost:8529/spline-test" with disabled authentication, and proceed with that assumption. For now rely on your local environment, we'll modify our CI pipeline later.

Thank you again for your efforts.

mohanajuhi166:
@wajda greetings .
Any chance if we can open this feature again?

wajda:
Sure, as soon as the PR is ready we'll review it again.

wajda:
Closed the PR because no work on it has been done for more than a year

Issue #805, opened 2020-11-25
Spline producer url unreachable from spark


siddharths067:
### Spark cannot connect to spline gateway, `HTTP Response 404`

I used the TLDR configuration and packages given in https://absaoss.github.io/spline/. `za.co.absa.spline.agent.spark:spark-2.4-spline-agent-bundle_2.12:0.5.5 and 'spark.spline.producer.url' as 'http://localhost:9090/producer'`

But I get this error in my spark job
```
20/11/25 04:05:03 ERROR QueryExecutionEventHandlerFactory: Spline initialization failed! Spark Lineage tracking is DISABLED.
za.co.absa.spline.harvester.exception.SplineInitializationException: Spark Agent was not able to establish connection to Spline Gateway. HTTP Response: 404 
```

The docker-compose logs showed this error in `spline` container
```
spline_1    | 04:15:29.924 [main] INFO  z.c.a.s.p.ArangoRepoConfig$$EnhancerBySpringCGLIB$$71b73625 - Connecting to arangodb://arangodb:8529/spline
spline_1    | 04:15:30.466 [main] WARN  z.c.a.s.g.r.AppInitializer$$anon$1 - Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'arangoRepoConfig': Invocation of init method failed; nested exception is java.lang.RuntimeException: Database version 0.4.0 is out of date, version 0.5.5 is required. Please execute 'java -jar admin-0.5.5.jar db-upgrade' to upgrade the database.
spline_1    | 04:15:30.482 [main] ERROR o.s.web.context.ContextLoader - Context initialization failed
spline_1    | org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'arangoRepoConfig': Invocation of init method failed; nested exception is java.lang.RuntimeException: Database version 0.4.0 is out of date, version 0.5.5 is required. Please execute 'java -jar admin-0.5.5.jar db-upgrade' to upgrade the database.
```

following the logs, I ran the recommended command by modifying the docker-compose image `java -jar admin-0.5.5.jar db-upgrade` as follows
```
    command: >
      bash -c "echo 'Initializing Spline DB...'
      && curl -O -s https://repo1.maven.org/maven2/za/co/absa/spline/admin/0.5.0/admin-0.5.0.jar
      && java -jar ./admin-0.5.0.jar db-init arangodb://arangodb/spline -s
[...]

PR #806, opened 2020-11-25
Spline #805 update docker compose

cerveada:
- update to the latest Spline version
- use fixed version to keep the compose process reproducible

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,18 +2,18 @@
 version: '3'
 services:
   arangodb:
-    image: arangodb:3.5.1
+    image: arangodb:3.7.1
     environment:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -2,26 +2,26 @@
 version: '3'
 services:
   arangodb:
-    image: arangodb:3.5.1
+    image: arangodb:3.7.1
```
wajda: Use Arangodb 3.7.3 due to arangodb/arangodb#12693

cerveada: We should update this on GitHub pages as well.

cerveada:
docker-compose.yml will be moved to different repo

Issue #810, opened 2020-12-02
Admin CLI can't deploy Foxx to a secured ArangoDB


wajda:
```
$ java -jar admin-0.5.5.jar  db-init arangodbs://aaa.bbb.ccc/spline -k

Exception in thread "main" org.apache.http.NoHttpResponseException: aaa.bbb.ccc:8529 failed to respond
	at org.apache.http.impl.conn.DefaultHttpResponseParser.parseHead(DefaultHttpResponseParser.java:141)
	at org.apache.http.impl.conn.DefaultHttpResponseParser.parseHead(DefaultHttpResponseParser.java:56)
	at org.apache.http.impl.io.AbstractMessageParser.parse(AbstractMessageParser.java:259)
	at org.apache.http.impl.DefaultBHttpClientConnection.receiveResponseHeader(DefaultBHttpClientConnection.java:163)
	at org.apache.http.impl.conn.CPoolProxy.receiveResponseHeader(CPoolProxy.java:157)
	at org.apache.http.protocol.HttpRequestExecutor.doReceiveResponse(HttpRequestExecutor.java:273)
	at org.apache.http.protocol.HttpRequestExecutor.execute(HttpRequestExecutor.java:125)
	at org.apache.http.impl.execchain.MainClientExec.execute(MainClientExec.java:272)
	at org.apache.http.impl.execchain.ProtocolExec.execute(ProtocolExec.java:186)
	at org.apache.http.impl.execchain.RetryExec.execute(RetryExec.java:89)
	at org.apache.http.impl.execchain.RedirectExec.execute(RedirectExec.java:110)
	at org.apache.http.impl.client.InternalHttpClient.doExecute(InternalHttpClient.java:185)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:83)
	at org.apache.http.impl.client.CloseableHttpClient.execute(CloseableHttpClient.java:108)
	at za.co.absa.spline.common.rest.RESTClientApacheHttpImpl.$anonfun$execHttp$5(RESTClientApacheHttpImpl.scala:72)
	at za.co.absa.commons.lang.ARM$.using(ARM.scala:30)
	at za.co.absa.commons.lang.ARM$ResourceWrapper.map(ARM.scala:41)
	at za.co.absa.spline.common.rest.RESTClientApacheHttpImpl.$anonfun$execHttp$4(RESTClientApacheHttpImpl.scala:72)
[...]

PR #811, opened 2020-12-03
spline #810 Admin CLI: support for REST over HTTPS

wajda:
depends on AbsaOSS/commons#44 (therefore a draft PR)


wajda:
Confirmed to be working with two types of certificates, locally as well as on AWS.
Please review.

The build is failing temporarily because of missing commons-0.0.19 in Maven central (they have a peak load issue atm, the sync hasn't complete yet)

persistence/src/main/scala/com/arangodb/async/internal/ArangoExecutorAsyncDestructor.scala
```diff
@@ -19,7 +19,7 @@ package com.arangodb.async.internal
 import com.arangodb.internal.velocystream.VstCommunication
 import za.co.absa.commons.reflect.ReflectionUtils
 
-object ArangoExecutorAsyncExtractor {
+object ArangoExecutorAsyncDestructor {
```
cerveada, line 22: What is the reason for this rename? Extractor seems ok, Destructor seems a bit confusing. To destruct usually means to destroy.

persistence/src/main/scala/za/co/absa/spline/persistence/ArangoDatabaseFacade.scala
```diff
@@ -51,7 +51,9 @@ class ArangoDatabaseFacade(connectionURL: ArangoConnectionURL) extends Disposabl
     arangoBuilder.build
   }
 
-  val db: ArangoDatabaseAsync = {
+  // The val should be lazy to not prevent a facade instance from being created.
+  // It would allow connection to be re-attempted later and the {{shutdown()}} method to be called.
+  lazy val db: ArangoDatabaseAsync = {
```
cerveada, line 56: I would simplify it: `The val is lazy ...` and `It allows connection ...`

persistence/src/main/scala/com/arangodb/async/internal/ArangoExecutorAsyncDestructor.scala
wajda: To me `FooExtractor` reads as it extracts Foo from something. While in fact in this case it actually extracts something from Foo. And that confused me when I looked at the code after a few months of writing it.

persistence/src/main/scala/za/co/absa/spline/persistence/ArangoDatabaseFacade.scala
wajda: done

persistence/src/main/scala/com/arangodb/async/internal/ArangoExecutorAsyncDestructor.scala
wajda: I was thinking of _destructurer_, _decomposer_... Unfortunately there is no naming convention or at least any recommendation how to name this thing in Scala world. In most code that I've seen it doesn't have its own as it's usually just an _unapply_ method in a companion object. Since I couldn't reuse the same naming approach for a library class and had do pick another name, I ended up with the word _destructor_ using vocabulary from this Kotlin related article - https://opencredo.com/blogs/the-destructor-pattern/

cerveada (approved):

Issue #814, opened 2020-12-10
Async producer API (Kafka)


wajda:
Add Karka producer API implementation, in addition to the current REST API.

Related to AbsaOSS/spline-spark-agent#166

wajda:
I think this should be another runtime component that is deployed separately from the gateway

PR #836, opened 2021-03-10
spline #814 kafka consumer for producer API

### LINKED ISSUE #839: Kafka consumer - error handling
Kafka give us a lot of options how to handle errors during message processing. Further analysis should be done. It would probably be good to make it configurable.

Some possible options:
- ack all messages and just log the errors
- push the failing message in special error topic and ack
- shutdown the application on error
- (other ?)

### LINKED ISSUE #840: Kafka consumer - performance imporvments
Spring provides `Batch listener` that allows accessing all messages in current batch. It would be possible to group the message by planId and then process them in parallel. 

Almost the same thing is theoretically possible on a higher level by deploying several Kafka servers and using Kafka partitioning, but I am not sure how the database and persistence layer will handle this.

cerveada:


producer-kafka/src/main/scala/com/twitter/finatra/FinatraInternalModules.scala
```diff
@@ -0,0 +1,29 @@
...
+import com.twitter.finatra.json.internal.caseclass.jackson.CaseClassModule
+
+/**
+  * This object serves the only purpose - to make access to some of the com.twitter.finatra Jackson modules,
+  * that are declared private even though are perfectly reusable.
+  */
+object FinatraInternalModules {
+
+  def caseClassModule: JacksonModule = CaseClassModule
+}
```
cerveada: Should this be moved to spline commons?

wajda: I'm not sure. This adapter is specific to `com.twitter:finatra-jackson` extension. I don't see any other project in AbsaOSS space that is using that library. So I don't think it's worth being promoted to Commons.

pom.xml
```diff
@@ -64,6 +64,8 @@
         <module>consumer-rest-core</module>
         <module>rest-gateway</module>
 
+        <module>producer-kafka</module>
```
wajda: what about renaming it to "kafka-gateway" for the sake of consistency?

producer-kafka/src/main/scala/com/twitter/finatra/FinatraInternalModules.scala
wajda: Ah, you mean Spline-Commons :) Then yes, with optional dependency.

producer-kafka/src/main/scala/za/co/absa/spline/producer/kafka/ProducerKafkaConfig.scala
```diff
@@ -0,0 +1,127 @@
...
+    val factory = new ConcurrentKafkaListenerContainerFactory[String, AnyRef]()
+    factory.setConcurrency(1)
+    factory.setConsumerFactory(consumerFactory)
+    factory.setMessageConverter(messageConverter)
+    factory.getContainerProperties.setAckMode(AckMode.MANUAL)
+    factory
+  }
+
+  private def consumerFactory: ConsumerFactory[String, AnyRef] = {
+    new DefaultKafkaConsumerFactory(consumerConfigsMerged.asJava);
```
wajda: nit: semicolon

producer-kafka/src/main/scala/za/co/absa/spline/producer/kafka/ProducerKafkaConfig.scala
```diff
@@ -0,0 +1,127 @@
...
+  import ProducerKafkaConfig._
+
+  @Bean
+  def kafkaListenerContainerFactory: ConcurrentKafkaListenerContainerFactory[_, _] = {
+    val factory = new ConcurrentKafkaListenerContainerFactory[String, AnyRef]()
+    factory.setConcurrency(1)
+    factory.setConsumerFactory(consumerFactory)
+    factory.setMessageConverter(messageConverter)
+    factory.getContainerProperties.setAckMode(AckMode.MANUAL)
+    factory
```
wajda: Let me suggest a slightly different code style for initialization blocks like this one. It's not necessarily (or universally) better (creates an anonymous subclass, so it's a subject to watch out), but is IMO visually clearer. Just a hint.
```suggestion
    new ConcurrentKafkaListenerContainerFactory[String, AnyRef] {
      setConcurrency(1)
      setConsumerFactory(consumerFactory)
      setMessageConverter(messageConverter)
      getContainerProperties.setAckMode(AckMode.MANUAL)
    }
```

producer-kafka/src/main/scala/za/co/absa/spline/producer/kafka/ProducerKafkaConfig.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  private val typeMappings = Map[String, Class[_]](
+    "ExecutionPlan" -> classOf[za.co.absa.spline.producer.model.v1_1.ExecutionPlan],
+    "ExecutionEvent" -> classOf[za.co.absa.spline.producer.model.v1_1.ExecutionEvent]
+  )
+
+  private val objectMapper: ObjectMapper =
+    JacksonUtils.enhancedObjectMapper()
+      .registerModule(DefaultScalaModule)
+      .setPropertyNamingStrategy(PropertyNamingStrategy.LOWER_CAMEL_CASE)
```
wajda: deprecated reference

producer-kafka/src/main/scala/za/co/absa/spline/producer/kafka/ProducerKafkaConfig.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  import za.co.absa.commons.config.ConfigurationImplicits._
+
+  val ConsumerKey = "spline.kafka.consumer"
+  val PlanTimeoutKey = "spline.kafka.insertPlanTimeout"
+  val EventTimeoutKey = "spline.kafka.insertEventTimeout"
+
+  val conf = this
+
+  val consumerConfig = ConfigurationConverter
```
wajda: public definitions require types

producer-kafka/src/main/scala/za/co/absa/spline/producer/kafka/ProducerKafkaConfig.scala
```diff
@@ -0,0 +1,127 @@
...
+    val sources = new MutablePropertySources()
+    sources.addLast(new PropertiesPropertySource("spline-property-source", properties))
+
+    val configurer = new PropertySourcesPlaceholderConfigurer()
+    configurer.setPropertySources(sources)
+    configurer
+  }
+}
+
+object ProducerKafkaConfig extends DefaultConfigurationStack {
```
wajda: Don't you want to use a typed config approach as is used in other modules?
For example:
```scala
object ProducerKafkaConfig extends DefaultConfigurationStack with ConfTyped {

  override val rootPrefix: String = "spline"
  private val conf = this

  object Kafka extends Conf("kafka") {
    val ConsumerConfig: Map[String, AnyRef] =
      getMap(subset(Prop("consumer")))
        .asScala.asInstanceOf[Map[String, AnyRef]]

    val PlanTimeout: Duration = conf
      .getOptionalLong(Prop("insertPlanTimeout"))
      .map(Duration(_, TimeUnit.MILLISECONDS))
      .getOrElse(1.minute)

    val EventTimeout: Duration = conf
      .getOptionalLong(Prop("insertEventTimeout"))
      .map(Duration(_, TimeUnit.MILLISECONDS))
      .getOrElse(10.seconds)
  }
}
```

producer-kafka/src/main/scala/za/co/absa/spline/producer/kafka/listener/ExecutionsListener.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  @KafkaHandler
+  def listenExecutionEvent(
+    event: ExecutionEvent,
+    ack: Acknowledgment,
+    @Header(KafkaHeaders.OFFSET) offset: Long,
+    @Header(KafkaHeaders.RECEIVED_TOPIC) topic: String,
+  ): Unit = {
+    try {
+      Await.result(repo.insertExecutionEvents(Array(event)), ProducerKafkaConfig.eventTimeout)
```
wajda: Ok for now, but as you pointed it out on the call there is a large room for optimization.
Basically we only need to block on `ExecPlan` messages, while `ExecEvent` ones can be processed in parallel as there is no relation between them.

cerveada: Yes, this will be covered by #840

cerveada:
At the end I decided to acknowledge all messages and just log the errors. It seems like a better default for now. 

More advanced error handling can be added later: https://github.com/AbsaOSS/spline/issues/839

wajda (approved):
lgtm

PR #841, opened 2021-03-10
Feature/spline 838 rest data sources

wajda:
Consumer REST API: add `GET /data-sources`

### DIFF
--- a/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/DataSourceRepositoryImpl.scala
+++ b/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/DataSourceRepositoryImpl.scala
@@ -1,0 +1,140 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/DataSourceRepositoryImpl.scala
```diff
@@ -0,0 +1,140 @@
...
+        Map(
+          "asAtTime" -> (asAtTime: java.lang.Long)
+        ))
+
+      val eventualArangoCursorAsync = db.queryAs[WriteEventInfo](
+        """
+          |WITH progress, progressOf, executionPlan, affects, dataSource
+          |FOR ds IN dataSource
+          |    FILTER ds._created <= @asAtTime
+          |    FILTER !LENGTH(@dataSourceUri) || @dataSourceUri == ds.uri
```
cerveada: Would just `!@dataSourceUri` work?

> null will be converted to false
an empty string is false, all other strings are true
https://www.arangodb.com/docs/stable/aql/operators.html

cerveada (dismissed):

cerveada (approved):

Issue #843, opened 2021-03-12
Consumer REST: Add "name" field for "DataSource" with sorting capabilities


wajda:
`GET /execution-events`
`GET /data-source`
should have sortable "name" field, that is derived from the "dataSourceURI" by cutting on the last meaningful slash

fyi, @oleksandr-korchak

oleksandr-korchak:
@wajda maybe it is better to call it `dataSourceName`?

wajda:
Makes sense for now, especially as both are sharing the same entity at the moment.

PR #845, opened 2021-03-12
spline #843 Consumer REST: Add "dataSourceName" field to the "WriteEventInfo" entity

wajda:
The field is derived from the URI as the last meaningful part of the URI


wajda:
@oleksandr-korchak

cerveada (approved):

PR #848, opened 2021-03-16
Bugfix/spline 835 Agent 0.5 (API 1.0) backward compatibility fix

wajda:


producer-model-mapper/src/main/scala/za/co/absa/spline/producer/modelmapper/v1/spark/SparkSplineOperationOutputConverter.scala
```diff
@@ -31,6 +31,7 @@ class SparkSplineOperationOutputConverter(
     attributeDefs
       .groupBy(_ (FieldNamesV1.AttributeDef.Id).toString)
       .mapValues(_.head)
+      .view.force // see: https://github.com/scala/bug/issues/4776
```
cerveada, line 34: This is just an idea.

Maybe we could define common `mapValueForce` function and use it everywhere as a workaround. So the problem would be documented on one place.

producer-rest-core/src/test/scala/za/co/absa/spline/producer/modelmapper/v1/spark/SparkSplineExpressionConverterSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+      )
+    ))
+
+    binExpr should be(a[FunctionalExpression])
+    binExpr.asInstanceOf[FunctionalExpression].id should not be empty
+    binExpr.asInstanceOf[FunctionalExpression].childIds should have length 2
+    binExpr.asInstanceOf[FunctionalExpression].name should equal("+")
+    binExpr.asInstanceOf[FunctionalExpression].dataType should equal(Some("d32cfc1f-e90f-4ece-93f5-15193534c855"))
+    binExpr.asInstanceOf[FunctionalExpression].params should be(empty)
+    binExpr.asInstanceOf[FunctionalExpression].extra should equal(Map(
```
cerveada, line 79: I would use variable instead of retyping all the time.

producer-rest-core/src/test/scala/za/co/absa/spline/producer/modelmapper/v1/spark/SparkSplineExpressionConverterSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  it should "support missing Literal.value" in {
+    val converter = new SparkSplineExpressionConverter(null)
+
+    val lit = converter.convert(Map("_typeHint" -> "expr.Literal"))
+
+    lit should be(a[Literal])
+    assert(lit.asInstanceOf[Literal].value == null)
+    assert(lit.asInstanceOf[Literal].dataType.isEmpty)
```
cerveada, line 53: Is there a reason why assert is used here and not should as everywhere else?

cerveada (approved):

PR #851, opened 2021-03-19
Feature/spline 789 db watermarks

wajda:
All persistent entities (including edges) that are scoped by a single execution plan to be marked with the ID of that plan. This allows for fast access to the aggregate and it's components without having to traverse the graph.

The implementation is done via introduce a new generic property `_parentId` to all persistent entities, that allows documents to be gathered in logical aggregates represented by other documents.


cerveada (commented):
Wouldn't it make more sense to name `_parentId` a `planId`?

- If evry parentId is planId why not name it that way in the first place?
- If some parentIds are not planIds why not name the ones that are planId planId to make it clear?

Another issue is that parent already have established meaning in oriented graphs. So it may be confusing.

wajda:
> Wouldn't it make more sense to name `_parentId` a `planId`?

I started with this and then quickly realized that a more generic solution would fit better. The problem that this property solves is not related to the domain, but rather is query optimization - helps to substitute graph traversals with simple filtering query by an indexed field. Ideally I wanted the change to be completely transparent to the majority of the code base, and be encapsulated on the deepest DAO layer. But I haven't come with a good solution in a short time frame, and had to let the field to appear in the persistent data model entities.
To at least somehow keep the logical distance between this property and the rest of properties with domain semantics I didn't name it after the _Execution Plan_, and used a generic name instead. Hence "Parent ID". It also allows for possible usage for something else in the future. I even thought to make it an array, so that documents could belong to multiple aggregates, but when such case appears a migration will be required anyway, so we could easily to this change later. And it wouldn't require property renaming nor AQL modification.

> Another issue is that parent already have established meaning in oriented graphs. So it may be confusing.

Ok, we could find a better name. But consider that:
- the property is defined in the `ArangoDocument` entity which is a super type for any kind of document that can be stored in the database, not just graphs.
- The term _parent_ in oriented graphs is used for describing a relation between nodes (which is represented by edges), not a node itself. So a property with such meaning would hardly appear in the node and create a logical collision with _parentId_

During my coding I tried names "_aggregateId" or "_belongsTo", but they just didn't sound to me.

cerveada:
- What I am a bit worried about is that when we have a multiple "aggregates" it will be hard to find you what parentId means for a given document type. There is no definition in the code it just depends on what you inserted.

- `_belongsTo` seems ok to me or `_ownerId`, or maybe `_ancestorId`

wajda:
> There is no definition in the code it just depends on what you inserted.

That's the intention - the meaning depends on how you query it, and it's always redundant from the domain model perspective.

Ok, will rename it to `_belongsTo`

cerveada (approved):

Issue #846, opened 2021-03-16
Rename package: za.co.absa.spline.producer.kafka` to `za.co.absa.spline.gateway.kafka` for consistency with the REST gateway


wajda:

cerveada:
It makes sense, but it's not 100% consistent either way. Rest-gateway provide bot producer and consumer APIs, whereas kafka-gateway provides only producer API.

wajda:
That's right, but those are different types of gateway. We could have one configurable gateway that could run an arbitrary number of services, like Kafka producer, REST producer and REST consumer. In that case Kafka module would only consist of Kafka specific classes and we'd call it like `producer-kafka-core` or something (similar to the modules that contain REST specific stuff).
But since we split it on the web container level, and Kafka module actually is a WAR that is similar to the rest-gateway in that regard, it makes more sense to call it `gateway` respectively. 
If you want to be picky, let's create two packages - one `gateway.kafka` is for keeping `AppInit` and `GatewayConfig` classes in there, and the other one `producer.kafka` to keep the listeners respectively :) Both in one module

wajda:
will be done in scope of #831

PR #860, opened 2021-04-09
Spline 831 Producer API enhancement

wajda:
- Renamed `childIds` to `childRefs` in expression and attributes
- Added optional `name` property to execution plan and operations
- Added `durationNs` property to the execution event
- Renamed package name in the `kafka-gateway` module (#846)
- Updated Producer REST API Swagger doc
- other minor fixes and refactoring


persistence/src/main/scala/za/co/absa/spline/persistence/model/operations.scala
```diff
@@ -17,43 +17,52 @@
 package za.co.absa.spline.persistence.model
 
 sealed trait Operation extends Vertex {
+  def name: Option[Operation.Name]
   def params: Map[String, Any]
   def extra: Map[String, Any]
-  def `type`: String
+  def `type`: Operation.Type
```
cerveada, line 23: What is the logic behind the naming? Sometime ``` `type` ``` and other time `_type`?

producer-model/src/main/scala/za/co/absa/spline/producer/model/v1_1/AttrOrExprRef.scala
```diff
@@ -46,15 +45,17 @@ case class AttrOrExprRef(
...
+  implicit class AttrOrExprOps(val ref: AttrOrExprRef) extends AnyVal {
+    def refId: ExpressionLike.Id = (ref.__attrId orElse ref.__exprId).get
 
-  def isExpression: Boolean = __exprId.isDefined
-}
+    def isAttribute: Boolean = ref.__attrId.isDefined
 
-object AttrOrExprRef {
+    def isExpression: Boolean = ref.__exprId.isDefined
+  }
```
cerveada, line 58: Implicits adds another complexity. Why to use them for this?

producer-model/src/main/scala/za/co/absa/spline/producer/model/v1_1/operations.scala
```diff
@@ -43,30 +45,32 @@ object OperationLike {
...
   append: Boolean,
   override val id: Id,
+  override val name: Option[OperationLike.Name] = None,
   override val childIds: Seq[Id],
   override val params: Map[String, Any] = Map.empty,
   override val extra: Map[String, Any] = Map.empty
 ) extends OperationLike {
-  override val output: Seq[Attribute.Id] = Nil
-
+  override def output: Seq[Attribute.Id] = Nil
```
cerveada, line 75: Why is def better?

persistence/src/main/scala/za/co/absa/spline/persistence/model/operations.scala
wajda: yes, a lot of confusion here... Originally the logic was to have all technical (non-business) props to be prefixed with underscore, just to avoid confusion in situations like when you add a business property with a similar name. For example operation type "Filter" vs type "transformation" or vs type "za.co.absa....operation.FilterOperation" -- here every type means different things and is supposed to be analyzed on different application layers. It worked well before and in general, but in case with Arango, since we don't have a control on how JSON fields are mapped on the class properties in Arango driver this convention leaked into the domain model and started making noise there. And also `_` is a reserved for Arango stuff, so I think we need to get rid of using that prefix convention.

producer-model/src/main/scala/za/co/absa/spline/producer/model/v1_1/AttrOrExprRef.scala
wajda: for proper serialization and Swagger generation. If you make them normal fields they will be propagated everywhere.

wajda: I tried `def` vs `val`, but it didn't help either. Implicits do the trick however.

producer-model/src/main/scala/za/co/absa/spline/producer/model/v1_1/operations.scala
wajda: Same as above -- swagger generation tweaking. It's quite sensitive on conventions, so the way you define stuff in the classes matters.

wajda: I think we'll need to switch to the Swagger definition first approach, and generate case classes from it. Otherwise it seems to be too fragile -- you add or change something in the code and don't realize that Swagger def is already different...

cerveada (dismissed):
LGTM

PR #865, opened 2021-04-12
spline #spline-864 Add a DB schema 0.6.0 version marker file

### LINKED ISSUE #864: Add a DB schema 0.6.0 version marker file
Implementation of the actual migration script (#673) can be postponed to the next patch release.

wajda:
fixes #864

### DIFF
--- a/persistence/src/main/resources/migration-scripts/0.6.0-0.6.0.js
+++ b/persistence/src/main/resources/migration-scripts/0.6.0-0.6.0.js
@@ -1,0 +1,19 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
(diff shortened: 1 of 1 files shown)


cerveada (approved):
OK, but why do we need the empty file? Does some code depend on it being present?

wajda:
The _current_ app DB schema version is automatically inferred from the script files provided. If this file isn't there, the server "thinks" that the required schema version is 0.5.5 (from the highest "to" version of the scripts), while in fact the persistent data model is different. This breaks the DB version checker and the incremental auto migration mechanism.
Putting just an empty file there completes the graph and makes things aligned. Even though there is no migration path from any other version to 0.6.0, it can be added in a further patch releases.

PR #875, opened 2021-04-22
Bugfix/spline 874 admin cli - fix ANSI codes rendering

wajda:


commons/src/main/scala/za/co/absa/spline/common/ConsoleUtils.scala
```diff
@@ -0,0 +1,50 @@
...
+      val ansiAllowed = System.console != null && System.getenv.get("TERM") != null
+      val ansiString  = $ansiInterpolation
+
+      if (ansiAllowed) ansiString
+      else stripAnsiEscapeSequences(ansiString)
+    """
+  }
+
+  def stripAnsiEscapeSequences(s: String): String = {
+    val sciSequencePattern = "\u001b\\[.*?[\\x40-\\x7E]"
```
cerveada:
```suggestion
    val ansiSequencePattern = "\u001b\\[.*?[\\x40-\\x7E]"
```
Is this typo?

cerveada (dismissed):

commons/src/main/scala/za/co/absa/spline/common/ConsoleUtils.scala
wajda: Ah, yes, it's a typo :)
Should be CSI (Control Sequence Introducer)
https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_(Control_Sequence_Introducer)_sequences

wajda:
```suggestion
    val csiSequencePattern = "\u001b\\[.*?[\\x40-\\x7E]"
```

wajda:
Tested in Linux as well as Windows -- standard terminal, PS and Cygwin (thanks @dk1844)

PR #880, opened 2021-04-26
added ARG for base_image and changed docker-entrypoint

dvagapov:
example of using NON-default base image:

`mvn dockerfile:build -Ddockerfile.tag=test -D**dockerfile.base.artifact**=openjdk -Ddockerfile.base.tag=11.0-jre-buster`

### DIFF
--- a/admin/Dockerfile
+++ b/admin/Dockerfile
@@ -31,10 +31,7 @@ LABEL \
     license="Apache License, version 2.0" \
     name="Spline Admin Tool"
 
-ENV ADMIN_COMMAND="db-init"
-ENV ARANGODB_URL="arangodb://localhost:8529/spline"
-ENV EXTRA_ARGS="-s"
-
 COPY target/admin-$VERSION.jar ./admin.jar
 
-ENTRYPOINT java -jar admin.jar $ADMIN_COMMAND $ARANGODB_URL $EXTRA_ARGS
+ENTRYPOINT ["java", "-jar", "admin.jar" ]
+CMD ["--help"]
--- a/admin/Dockerfile
+++ b/admin/Dockerfile
@@ -13,10 +13,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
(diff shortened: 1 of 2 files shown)


admin/Dockerfile
```diff
@@ -13,18 +13,25 @@
...
     license="Apache License, version 2.0" \
     name="Spline Admin Tool"
 
-COPY target/admin-$version.jar ./admin.jar
+COPY target/admin-$VERSION.jar ./admin.jar
 
-ENTRYPOINT ["java", "-jar", "admin.jar"]
-CMD ["--help"]
+ENTRYPOINT ["java", "-jar", "admin.jar" ]
+CMD ["--help"]
```
cerveada, line 37: Missing line end

admin/pom.xml
```diff
@@ -98,9 +100,12 @@
                 <configuration>
                     <skip>false</skip>
                     <buildArgs>
-                        <version>${project.version}</version>
+                        <DOCKER_BASE_ARTIFACT>${dockerfile.base.artifact}</DOCKER_BASE_ARTIFACT>
```
Zejnilovic, line 103: How about adding `<JAR_FILE>${project.build.finalName}.jar</JAR_FILE>` to build args and in Dockerfile have `ARG JAR_FILE` instead of `ARG VERSION` and on `COPY` you use that one.

This would make it also easy to build docker images outside of this mvn context.

wajda:
@Zejnilovic, that's a tempting logical approach, but I used to have issues with the `finalName` not being respected equally in different use cases. Specifically Sonatype plugin didn't respect finalName on WAR artifacts. So I'd prefer stability and predictability over DRY.

admin/pom.xml
Zejnilovic: @wajda sure thing, if you find it problematic/unstable for your use case.

wajda:
I'll merge it, and the work will continue in the other PR

PR #888, opened 2021-05-03
spline #887 Switch to a released package-pom

### LINKED ISSUE #887: Switch to a released package-pom
`package-pom` has been move to a separate repo - https://github.com/AbsaOSS/spline-root-pom
Delete one from this repo, and fix references accordingly.

wajda:
fixes #887


build/parent-pom/pom.xml
```diff
@@ -21,13 +21,14 @@
     <parent>
         <groupId>za.co.absa.spline</groupId>
         <artifactId>package-pom</artifactId>
```
cerveada, line 23: Shouldn't the name be `spline-root-pom`?

wajda: As it might be promoted to the ABSA OSS level and hence eventually evicted from the Spline group of artifacts, I thought it would be better to keep the artifact name unchanged to not pollute the [group name space](https://repo1.maven.org/maven2/za/co/absa/spline/) with another short living name.

cerveada (dismissed):

cerveada (approved):

Issue #915, opened 2021-06-25
Kafka Gateway: improve configuration


wajda:
### Current state
In the `KafkaGatewayConfig` there is nothing tells that `spline.kafka.topic`, `spline.consumer.group.id ` or `spline.consumer.bootstrap.servers` are mandatory properties.
If one of more of them isn't provided different kind of exceptions are thrown from different parts of Kafka driver, and the messages are mostly misleading.
For instance:
```
Caused by: org.apache.kafka.common.config.ConfigException: No resolvable bootstrap urls given in bootstrap.servers
```
```
Caused by: java.lang.IllegalStateException: No group.id found in consumer config, container properties, or @KafkaListener annotation; a group.id is required when group management is used.
```
```
Caused by: java.lang.IllegalArgumentException: Could not resolve placeholder 'spline.kafka.topic' in value "${spline.kafka.topic}"
```

### Desired state
All required properties are explicitly defined in the `KafkaGatewayConfig`, and the error messages are straightforward and consistent:
```
Caused by: java.util.NoSuchElementException: Missing configuration property spline.kafka.topic
```
```
Caused by: java.util.NoSuchElementException: Missing configuration property spline.consumer.group.id
```
```
Caused by: java.util.NoSuchElementException: Missing configuration property spline.consumer.bootstrap.servers
```

PR #916, opened 2021-06-25
spline #915 Kafka Gateway: improve configuration

wajda:
- Make required config properties explicit
- Make missing properties error messages consistent
- Read `topic` property value from the same config stack as for the rest properties


cerveada (approved):
LGTM

PR #917, opened 2021-06-28
spline #834 Ensure SSL validation is supported

### LINKED ISSUE #834: Ensure SSL validation is supported
Per discussion with @dvagapov, we need to make sure that all our components that connect via secured protocols to external parties (ArangoDB driver, Kafka consumer etc) support TLS validation.
Also the respective Docker containers should expose options to provide CA cert to the container.

wajda:
fixes #834 

- Conditionally create AllTrusting SSL context (only if disabling SSL validation is explicitly requested via a corresponding configuration property or CLI argument)
- Make `-k` argument completely optional
- Add `--disable-ssl-validation` and deprecate `-k`, `--insecure`

### DIFF
--- a/admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
+++ b/admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
@@ -23,6 +23,7 @@ import scopt.{OptionDef, OptionParser}
 import za.co.absa.spline.admin.AdminCLI.AdminCLIConfig
 import za.co.absa.spline.common.ConsoleUtils._
 import za.co.absa.spline.common.SplineBuildInfo
+import za.co.absa.spline.common.security.TLSUtils
 import za.co.absa.spline.persistence.AuxiliaryDBAction._
 import za.co.absa.spline.persistence.OnDBExistsAction.{Drop, Fail, Skip}
 import za.co.absa.spline.persistence.{ArangoConnectionURL, ArangoManagerFactory, ArangoManagerFactoryImpl}
@@ -37,7 +38,7 @@ object AdminCLI extends App {
   case class AdminCLIConfig(
     cmd: Command = null,
     logLevel: Level = Level.INFO,
-    untrusted: Boolean = false,
+    disableSslValidation: Boolean = false,
   )
 
   implicit class OptionParserOps(val p: OptionParser[AdminCLIConfig]) extends AnyVal {
@@ -89,10 +90,10 @@ class AdminCLI(dbManagerFactory: ArangoManagerFactory) {
           action ((str, conf) => conf.copy(logLevel = Level.valueOf(str))))
       }
 
-      // FIXME: rename 'insecure' to 'untrusted'. See https://github.com/AbsaOSS/spline/issues/906
+      // FIXME: rename 'insecure' to 'disable-ssl-validation'. See https://github.com/AbsaOSS/spline/issues/906
       (opt[Unit]('k', "insecure")
         text s"Allow untrusted server connections when using SSL; disallowed by default."
-        action { case (_, conf) => conf.copy(untrusted = true) })
+        action { case (_, conf) => conf.copy(disableSslValidation = true) })
 
       this.placeNewLine()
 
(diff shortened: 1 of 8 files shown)


admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
```diff
@@ -166,23 +165,27 @@ class AdminCLI(dbManagerFactory: ArangoManagerFactory) {
       .asInstanceOf[Logger]
       .setLevel(conf.logLevel)
 
+    val sslCtxOpt =
+      if (!conf.disableSslValidation) None
+      else Some(TLSUtils.TrustingAllSSLContext)
```
cerveada:
```suggestion
      if (conf.disableSslValidation) Some(TLSUtils.TrustingAllSSLContext)
      else None
```

admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
```diff
@@ -89,10 +90,10 @@ class AdminCLI(dbManagerFactory: ArangoManagerFactory) {
           action ((str, conf) => conf.copy(logLevel = Level.valueOf(str))))
       }
 
-      // FIXME: rename 'insecure' to 'untrusted'. See https://github.com/AbsaOSS/spline/issues/906
+      // FIXME: rename 'insecure' to 'disable-ssl-validation'. See https://github.com/AbsaOSS/spline/issues/906
       (opt[Unit]('k', "insecure")
         text s"Allow untrusted server connections when using SSL; disallowed by default."
-        action { case (_, conf) => conf.copy(untrusted = true) })
+        action { case (_, conf) => conf.copy(disableSslValidation = true) })
```
cerveada: Why not support both keys for now, and remove the deprecated one later?

admin/src/test/scala/za/co/absa/spline/admin/AdminCLISpec.scala
```diff
@@ -93,6 +96,17 @@ class AdminCLISpec
       } should include("--help")
     }
 
+    it should "when calling with option -k, create a non-validating SSLContext" in {
```
cerveada:
```suggestion
    it should "create a non-validating SSLContext when called with option -k, " in {
```
Not important, but just bit nicer to read.

persistence/src/main/scala/za/co/absa/spline/persistence/ArangoRepoConfig.scala
```diff
@@ -33,7 +34,13 @@ class ArangoRepoConfig extends InitializingBean with Logging {
     log.info(s"Spline database URL: ${Database.ConnectionURL.asString}")
   }
 
-  @Bean def arangoDatabaseFacade: ArangoDatabaseFacade = new ArangoDatabaseFacade(Database.ConnectionURL)
+  @Bean def arangoDatabaseFacade: ArangoDatabaseFacade = {
+    val sslCtxOpt =
+      if (!Database.DisableSSLValidation) None
+      else Some(TLSUtils.TrustingAllSSLContext)
```
cerveada: Unnecessary negation

wajda: OK, will flip it.
The reason why I added extra negation in those two `if`'s was just to keep code lines shorter.

admin/src/test/scala/za/co/absa/spline/admin/AdminCLISpec.scala
wajda: Changed to "called", but I would leave the phase arrangement as is because it's consistent with the rest of "should" statements in the class. And it also visually groups related statements in test logs and in the code.

image

persistence/src/main/scala/za/co/absa/spline/persistence/ArangoRepoConfig.scala
wajda: I've added `Option` builders from Scala 13 and used them instead of `if-else` block. I think it reads even better.
```scala
val sslCtxOpt = Option.when(conf.disableSslValidation)(TLSUtils.TrustingAllSSLContext)
```

admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
cerveada: Or do you want to have the renaming in separated PR?

admin/src/test/scala/za/co/absa/spline/admin/AdminCLISpec.scala
cerveada: Since flatspec is going for "english sentence like" names, I would personally change all of them to feel like natural language, but I can live with the current version.

admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
wajda: I planned to do it in #906, but I think we can have both keys as you said. Will add one

cerveada (approved):
LGTM

PR #922, opened 2021-06-30
spline #673 DB migration from v.0.5.5 to 0.6

wajda:
The migration takes care of most model parts, excepts for expressions and attribute dependencies. The later was omitted due to complexity and lack of time to implement it. This effectively makes the 0.5 to 0.6 migration not 100% complete, but since expressions and attribute level lineage in 0.5 was experimental feature and only partially implemented, sacrificing those details shouldn't be an issue for majority of 0.5 users.


wajda:
@cerveada I tested it on a small databases, it seems to be working. But it also needs to be tested on bigger ones, e.g. Jedi.
I tried to migrate Jedi 0.4.0 dump on the localhost, but it fails with "no space left on device". It seems to consume RAM and disk aggressively. I wonder how would it behave if it's given more resources. I'm not sure if it's worth extra work on optimizing it (if there are easy ways to optimize). If it's greedy but works I would release it.
Could you please run some tests? Thanks.

cerveada:
Code seems to be OK.
Testing is still in progress.

cerveada:
To test this, I created two deployments:
- 0.5.6 Agent, Gateway and ArangoDB 
- 0.6  Agent, Gateway and ArangoDB

Using the agent's examples I generated some data, migrated 0.5 DB to 0.6 using this PR and compared both in two UIs

Seems to be working on level of plan and operations. The attribut lineage is missing as expected, but there some additional artefacts: (migrated version on the left)

Screenshot 2021-07-01 at 14 45 25
This is of course caused by better union handling in 0.6. I don't know if it's a bug, but it should be considered.

Screenshot 2021-07-01 at 14 52 52
This seems to be a bug? Why is there so much more attributes in the migrated version's input?

wajda:
In both cases the difference is caused by agents. not by migration.

wajda:
I've double checked, the difference in the seen `nasa.xml` schema is caused by the line in the `ExampleJob3.scala` below, that was added in Agent 0.6.

image

cerveada (dismissed):

cerveada (approved):

PR #929, opened 2021-07-20
spline #928 Consumer: add filter by write mode

### LINKED ISSUE #928: Consumer: add filter by write mode
`GET .../execution-events/?...&append=true|false`

wajda:
fixes #928

- Added optional `append` query parameter to `/execution-events` and `/data-sources` routes.
Possible values are `true | false | 1 | 0 | undefined`
- Refactoring: 
  - replaced `NOT x` with `IS_NULL(x)` in AQL queries
  - replaced Scala type ascription with explicit boxing/unboxing where appropriate
  - converted nullable `T` to `Option[T]` where appropriate

### DIFF
--- a/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/DataSourceRepositoryImpl.scala
+++ b/consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/DataSourceRepositoryImpl.scala
@@ -38,9 +38,10 @@ class DataSourceRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends Dat
     writeTimestampEnd: Long,
     pageRequest: PageRequest,
     sortRequest: SortRequest,
-    searchTerm: String,
-    writeApplicationId: String,
-    dataSourceUri: String
+    maybeSearchTerm: Option[String],
+    maybeAppend: Option[Boolean],
+    maybeWriteApplicationId: Option[String],
+    maybeDataSourceUri: Option[String]
   )(implicit ec: ExecutionContext): Future[(Seq[WriteEventInfo], Long)] = {
 
     db.queryAs[WriteEventInfo](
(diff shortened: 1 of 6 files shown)


consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/DataSourceRepositoryImpl.scala
```diff
@@ -96,16 +96,17 @@ class DataSourceRepositoryImpl @Autowired()(db: ArangoDatabaseAsync) extends Dat
...
-        "asAtTime" -> (asAtTime: java.lang.Long),
-        "timestampStart" -> (writeTimestampStart: java.lang.Long),
-        "timestampEnd" -> (writeTimestampEnd: java.lang.Long),
-        "pageOffset" -> (pageRequest.page - 1: Integer),
-        "pageSize" -> (pageRequest.size: Integer),
+        "asAtTime" -> Long.box(asAtTime),
+        "timestampStart" -> Long.box(writeTimestampStart),
+        "timestampEnd" -> Long.box(writeTimestampEnd),
+        "pageOffset" -> Int.box(pageRequest.page - 1),
+        "pageSize" -> Int.box(pageRequest.size),
```
cerveada, line 103: Just curious: Why is the boxing needed here?

cerveada (approved):

consumer-services/src/main/scala/za/co/absa/spline/consumer/service/repo/DataSourceRepositoryImpl.scala
wajda: Because the `db.queryAs()` method expects `Map[String, AnyRef]`, but when you put a value of a primitive type to the map, the type is inferred as `Map[String, Any]` instead. Previously I've used type ascription to induce an auto boxing, but now it occurs to me that explicitly calling `box()` method on a respective type companion object is cleaner and more readable. Especially in other places, e.g. when doing `Option.map()` for example:
```scala
Option(42).map(x => x : java.lang.Integer)
// versus
Option(42).map(Int.box)
```

PR #930, opened 2021-07-29
spline #925 add unique constraint validation

wajda:
Add unique constraint validation on ExecutionPlan operation IDs, attribute IDs and expression IDs

### DIFF
--- a/commons/src/main/scala/za/co/absa/spline/common/validation/ValidationUtils.scala
+++ b/commons/src/main/scala/za/co/absa/spline/common/validation/ValidationUtils.scala
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2021 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


commons/src/main/scala/za/co/absa/spline/common/validation/ValidationUtils.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  @throws[ConstraintViolationException]
+  def validate(c: Constraint): Unit = {
+    validate(c.isValid, c.message)
+  }
+
+  @throws[ConstraintViolationException]
+  def validate(b: Boolean, msg: => String): Unit = {
+    if (!b) throw new ConstraintViolationException(msg, null) // NOSONAR
+  }
```
cerveada, line 31: - I don't understand why is this separated method. Could both validate methods be just merged into one?
- the name doesn't describe what the method does. Maybe `throwIfNotValid` would make more sense

wajda: It's a utility object that aims for the samilar purpose as JSR-380, but in a simplistic implementation without annotations and external dependencies. The method is called `validate()` as the standard `Validator.validate()`, and it's overloaded to support simple checks that don't necessarily require a dedicated constraint class. E.g.
```scala
ValidationUtils.validate(age > 0, s"Invalid age: $age")
```

cerveada: ok, used like that it makes more sense

cerveada (approved):
LGTM

Issue #939, opened 2021-08-09
Slow performance of the Dataset list query


wajda:
- missing index on `dataset._created`
- `"dataSourceName": REGEX_MATCHES(...)` can be calculated on insert and accompanied with an index.

PR #940, opened 2021-08-09
spline #939 Slow performance of the Dataset list query

wajda:
- add `dataSource.name` field + index
- add `progress.execPlanDetails.dataSourceName` + index
- add index `dataSource._created`
- add migration script
- rename `0.6.2-SNAPSHOT` to `0.7.0-SNAPSHOT` version (and the branch) due to database schema changes.

### DIFF
--- a/persistence/src/main/scala/za/co/absa/spline/persistence/model/entities.scala
+++ b/persistence/src/main/scala/za/co/absa/spline/persistence/model/entities.scala
@@ -95,7 +95,7 @@ object DataSource {
   type Uri = String
   type Name = String
 
-  private val NameRegexp = "(?:.*/)?([^/]+)/*".r
+  private val NameRegexp = "([^/]+)/*$".r
 
   def getName(uri: Uri): Name = {
     NameRegexp.findFirstMatchIn(uri).map(_.group(1)).getOrElse("")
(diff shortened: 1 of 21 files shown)


persistence/src/main/scala/za/co/absa/spline/persistence/model/entities.scala
```diff
@@ -84,14 +84,22 @@ object DBVersion {
...
-  def this() = this(null, null)
+  def this() = this(null, null, null)
 }
 
 object DataSource {
   type Key = ArangoDocument.Key
   type Uri = String
+  type Name = String
+
+  private val NameRegexp = "(?:.*/)?([^/]+)/*".r
```
cerveada: Why are the regexes here and in the script different?

wajda: I just thought it was clearer to explicitly specify a non-capturing group to make the whole string match, not just a suffix. Essentially it makes no difference in the result but IMO more readable. But I can use the same one if you prefer.

cerveada: But if it's more readable, I would use it in the script as well.

By the way, for me, more readable is the one in the script. It's easier to see it will take the part at the end, here I have to consider the whole string.

Either way, I think it's better to use the same one on both places.

cerveada (dismissed):
As I wrote in a comment, I would use the same regexp on both places, the rest LGTM

cerveada (approved):

PR #941, opened 2021-08-11
Bugfix/spline 931 loosen output schema consistency check

wajda:
The PR makes `Operation.output` optional. This makes it possible to distinguish between the _empty_ and _unspecified_ output.
The later is eligible for schema inference, while the former represents a _unit_ relation.

Also updated `rest-api-doc-generator` to the bugfixed version 1.0.4


cerveada:
So, this is and API change. Shouldn't there be a model 1.2?

wajda:
It's a DTO mapping change. The API itself is the same except for a minor change in the semantics of the empty attribute list that is now not eligible for inference. But the empty list semantics wasn't specifically mentioned in the documentation, and used to be mostly silently treated as an equivalent to unspecified schema (the semantics of which wasn't changed). From this perspective I don't think we break any documented convention on the API level.

cerveada (approved):

Issue #969, opened 2021-10-04
Consumer API: Data Source filters are counter-intuitive 


wajda:
Currently the _Last Executed_ relation is __outer__ joined to the _Data Source_. It usually makes sense unless at least one of the _Execution_ related filters is applied to the query. In that case the logic should correspond to an __inner join__ instead.

For example, when I as a user enable the `datetime` filter I expect to only see those data sources whose associated _Last Executed_ event satisfies the given date time criteria. The same holds for other _Execution_ related filters like `writeMode` for instance.

PR #972, opened 2021-10-04
Feature/spline 971 timestamp range flag

### LINKED ISSUE #971: Do not calculate timestamp range unnecessarily
Consumer API requests calculate _total timestamp range_ facet unconditionally. The client (UI) however doesn't need it for certain types of frequent queries (e.g. polling the server for updates). To avoid wasting resources on this expensive calculation we need to add a query parameter that would enable _total timestamp range_ facet on demand.

wajda:
- Added `facet.timestamp` boolean request param to conditionally enable total timestamp facet computation
- Made `timestampRange` optional (for consistency with #969)
- Added `DataSourceRepository.getTimestamp()` method that calculates the facet more correctly than `ExecutionEventRepository.getTimestamp()` method.


cerveada:
> Added DataSourceRepository.getTimestamp() method that calculates the facet more correctly than ExecutionEventRepository.getTimestamp() method.

I am bit confused about this:

- I guess you meant `getTimestampRange`?
- Why not used only the DataSourceRepository's one when it is more correct?

wajda:
`getTimestampRange` yes, that's a typo.

Although technically (at the moment) both use cases use the same data structures, there is a small difference in semantics. The execution event list represents all events, while the data source list is grouped by URI taking the latest event in a group plus data sources with no associated events. Hence the later cannot really reuse the query designed for the former. Even though it worked, the found range for the DataSource list sometimes started on earlier dates (that's what I meant by saying that it wasn't quite correct application for the method). I didn't mean the original method wasn't correct though. It was, for the ExecutionEvent list.

cerveada (dismissed):

cerveada (approved):

PR #986, opened 2021-10-22
Feature/spline 978 UUID collision handling

### LINKED ISSUE #978: ExecutionPlan UUID collision handling
This is an extreme precaution to remedy for the practically improbable but still theoretically possible UUID collision.

In the _Producer Model_, we could have a special property (like a checksum but selected by the user) that would serve as a business identifier, and would be included in both _Execution Plan_ and _Execution Event_ entities. The possible collision then could be detected by comparing that property with the previously stored data having the same UUID, and _if_ it happens to be different the document would be stored in a special log or a collection for manual handling.
In my opinion, giving an extremely small probability of the collision in the first place such treatment is way more than enough to make sure we won't loose any lineage from any job ever.

wajda:
- POM: upgraded _ArangoDB Java driver_ and _Logback_ dependencies
- Introduced `ExecutionPlan.discriminator` string property, which is additionally verified on every plan or event insert. If IDs of two entities match, but discriminators don't we conclude detected collision.

### DIFF
--- a/producer-model/src/main/scala/za/co/absa/spline/producer/model/v1_1/executionPlan.scala
+++ b/producer-model/src/main/scala/za/co/absa/spline/producer/model/v1_1/executionPlan.scala
@@ -21,8 +21,9 @@ import za.co.absa.spline.common.validation.{Constraint, ValidationUtils}
 import java.util.UUID
 
 case class ExecutionPlan(
-  id: UUID = UUID.randomUUID(),
+  id: ExecutionPlan.Id = UUID.randomUUID(),
   name: Option[ExecutionPlan.Name],
+  discriminator: Option[ExecutionPlan.Discriminator] = None,
 
   operations: Operations,
   attributes: Seq[Attribute] = Nil,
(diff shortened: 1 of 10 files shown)


producer-model/src/main/scala/za/co/absa/spline/producer/model/v1_1/executionPlan.scala
```diff
@@ -21,8 +21,9 @@ import za.co.absa.spline.common.validation.{Constraint, ValidationUtils}
 import java.util.UUID
 
 case class ExecutionPlan(
-  id: UUID = UUID.randomUUID(),
+  id: ExecutionPlan.Id = UUID.randomUUID(),
```
cerveada, line 24: Shouldn't the id value come only from the agent?

cerveada (approved):
lgtm, but I added one question

producer-model/src/main/scala/za/co/absa/spline/producer/model/v1_1/executionPlan.scala
wajda: Producer API v1.1 allows for the old round-trip communication style when the client sends a plan (without ID), the server generates one and returns it to the client. It's done by that default value.

PR #999, opened 2021-12-30
Feature/spline 737 labels

### LINKED ISSUE #737: Semi multi-tenancy support
The system should allow for metadata isolation based on different criteria such as:
  - the source system
  - function unit
  - project name
  - {{TBD}}

For now the isolation should be done logically, keeping the data physically on a shared persistent storage, maintaining links and data consistency. The goal is to isolate the users form the "noisy neighbors" on the UI and Consumer API level, while preserving ability to query the global lineage graph as a whole (e.g. for performing end-to-end lineage, impact or other analysis across the boundary of particular groups of users, projects, business units etc)

Security aspects are not considered in the scope of this epic.

Example user story:
> As a user I want to be able to filter/group entities based on custom tags provided by agents. (e.g. team, business unit etc)

wajda:
- Added `labels` property to both `ExecutionPlan` and `ExecutionEvent` entities
- Added scrollable REST for all labels and their values (`/labels/names/` and `/labels/names/{label_name}/values/` respectively)
- Added filters for `/execution-events` and `/data-source` resources to filter by labels.

Filter query format:
```
?label=aaa:bbb,ccc&label=xxx:yyy,zzz
```
is understood as
```
WHERE labels['aaa'] CONTAINS ANY OF ['bbb', 'ccc']
  AND labels['xxx'] CONTAINS ANY OF ['yyy', 'zzz']
```
Label names and values are case-sensitive.


cerveada (approved):
LGTM

PR #1006, opened 2022-01-20
(minor) Make Admin tool more talkative

wajda:
Change log level of some messages from DEBUG to INFO to make the output slightly more verbose. For instance, create/drop collections, index, service etc.

### DIFF
--- a/persistence/src/main/scala/za/co/absa/spline/persistence/ArangoManager.scala
+++ b/persistence/src/main/scala/za/co/absa/spline/persistence/ArangoManager.scala
@@ -18,6 +18,7 @@ package za.co.absa.spline.persistence
 
 import com.arangodb.async.ArangoDatabaseAsync
 import com.arangodb.entity.{EdgeDefinition, IndexType}
+import com.arangodb.model.Implicits.IndexOptionsOps
 import com.arangodb.model._
 import org.apache.commons.io.FilenameUtils
 import org.slf4s.Logging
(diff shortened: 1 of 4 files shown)


persistence/src/main/scala/za/co/absa/spline/persistence/ArangoManager.scala
```diff
@@ -127,18 +128,21 @@ class ArangoManagerImpl(
...
         throw new IllegalArgumentException(s"Arango Database ${db.name()} already exists")
       else if (exists && dropIfExists) {
-        log.debug("Drop existing database")
+        log.info(s"Drop database: ${db.name}")
         db.drop().toScala
       }
-      else Future.successful(Unit)
-    } yield Unit
+      else Future.successful({})
+    } yield {}
```
cerveada, line 135: In my opinion, `{}` instead of `Unit` is making the code harder to understand.

cerveada (approved):

persistence/src/main/scala/za/co/absa/spline/persistence/ArangoManager.scala
wajda: The reason why I removed `Unit` is because it is _not_ an instance of the type `Unit`. The only instance of it is `()`. So the proper writing of the end of that `for` block would be one of these: `yield ()`, `yield {()}` or `yield {}`

cerveada: OK, so it's correct. Still, it's a bit confusing. There is a `Future.unit` but only in Scala 2.12:
https://www.scala-lang.org/api/2.12.15/scala/concurrent/Future$.html#unit:scala.concurrent.Future[Unit]

Let's keep it as it is for now.

PR #1010, opened 2022-01-27
spline #1009 Remove obsolete producer model v1.1

### LINKED ISSUE #1009: Remove obsolete producer model v1.1

wajda:
- rename `producer-model-v1.1` module to `producer-model-v1.2`
- remove Producer model 1.1 related source code from the project, and replace it with a dependency on an external artifact.
- update related POMs accordingly


kafka-gateway/pom.xml
```diff
@@ -46,14 +46,28 @@
...
             <groupId>za.co.absa.spline</groupId>
-            <artifactId>producer-model-mapper</artifactId>
+            <artifactId>producer-model-v1.2</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <!-- deprecated models -->
+
+        <dependency>
+            <groupId>za.co.absa.spline</groupId>
```
AdrianOlosutean, line 64: Why still keeping this?

AdrianOlosutean (dismissed):

cerveada:
The build is falling
```
Could not find artifact za.co.absa.spline:producer-model-v1.1:jar:1.0.0-SNAPSHOT
```
Looks like there is still wrong dependecny somewhere?

kafka-gateway/pom.xml
wajda: Both Kafka and REST endpoints of Spline Producer API provide backward compatibility with some of the older models via the model mapper. The concrete models are defined as optional dependencies for the mapper, thus the client modules have to explicitly declare what models they actually support.

cerveada (approved):

PR #1012, opened 2022-01-27
spline #1011 test data generator module

### LINKED ISSUE #1011: Add Test Data Generator Module
## Background
In order to conduct performance and stress testing for Spline Producer API, a sample data generator is needed.

## Feature
Create the module for generating testing data for Spline producer

AdrianOlosutean:
Closes #1011 
Added module for generating data for spline producer


test-data-generator/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>za.co.absa.spline</groupId>
+            <artifactId>producer-model-v1.1</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
```
wajda: Unnecessary reference to a transitive dependency.

test-data-generator/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+        <dependency>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+        </dependency>
```
wajda: Unnecessary reference to a transitive dependency.

test-data-generator/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-library</artifactId>
+            <version>${scala.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-reflect</artifactId>
+            <version>${scala.version}</version>
+        </dependency>
```
wajda: Unnecessary reference to a transitive dependency.

test-data-generator/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+        <dependency>
+            <groupId>za.co.absa.spline</groupId>
+            <artifactId>commons</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.scala-lang</groupId>
+            <artifactId>scala-library</artifactId>
+            <version>${scala.version}</version>
+        </dependency>
```
wajda: A duplicate of an inherited dependency

test-data-generator/src/main/scala/EventGenerator.scala
```diff
@@ -0,0 +1,30 @@
...
+
+object EventGenerator {
+
+  def generate(executionPlan: ExecutionPlan): ExecutionEvent = {
+    ExecutionEvent(
+      planId = executionPlan.id,
+      timestamp = System.currentTimeMillis(),
+      durationNs = None,
+      error = None,
+      extra = Map()
```
wajda: nit: `Map.empty` is more conventional FP-style notation, while `Map()` looks more natural for imperative/mutable style as it resembles creation of a new instance of `Map'.

test-data-generator/src/main/scala/App.scala
```diff
@@ -15,19 +15,36 @@
...
+  def main(args : Array[String]): Unit = {
+    import scopt.OParser
+    val builder = OParser.builder[Config]
+    val configParser = {
+      import builder._
+      OParser.sequence(
+        programName("test-data-generator"),
+        head(
+          s"""
+             |Spline Data Generator
```
wajda:
```suggestion
             |Spline Test Data Generator
```

test-data-generator/src/test/java/PlanGeneratorTest.scala
```diff
@@ -0,0 +1,55 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import org.scalatest.funsuite.AnyFunSuite
+import za.co.absa.spline.producer.model.v1_2.{DataOperation, ExecutionEvent}
+
+class PlanGeneratorTest extends AnyFunSuite {
```
wajda: In Spline, we rather prefer using BDD style for writing test. I'm not against `AnyFunSuite` per se, but `AnyFlatSpec` combined with `should.Matchers` offers more natural DSL for writing BDD style tests. Consider using that one. And instead of testing methods (`generateWrite`, `generateRead`) focus on cases (`it should "generate one Write and two Reads when this and that condition is given" in { ... }"`).

wajda (changes requested):
In addition to the other comments, please move all the module source code to a corresponding package, say `za.co.absa.spline.testdatagen`. Also the directory `src/test/java` should be renamed to `src/test/scala` since it contains Scala files.

wajda:
Also, `send-plan.sh` why is it located in the `resources`? Those are meant rather for runtime usage and by default are copied into the target jar-file. That shell script is not a source code, so I would put it somewhere outside the `src/` folder, e.g. right into the root module directory, or a `bin/` directory if you wish.

wajda:
Sorry for flooding with comments, since in this PR we are creating a new module that is meant to be a CLI tool it would make sense to finish the POM part accordingly. Could you please add a `maven-assembly-plugin` to build an executable final JAR artifact. (See the `admin/pom.xml` for example).

test-data-generator/pom.xml
```diff
@@ -59,4 +59,33 @@
...
+                <artifactId>maven-assembly-plugin</artifactId>
+                <version>3.3.0</version>
+                <configuration>
+                    <descriptorRefs>
+                        <descriptorRef>jar-with-dependencies</descriptorRef>
+                    </descriptorRefs>
+                    <appendAssemblyId>false</appendAssemblyId>
+                    <archive>
+                        <manifestEntries>
+                            <Main-Class>za.co.absa.spline.splinegen.App</Main-Class>
```
wajda: Package name is incorrect.
Also let's make the main class name consistent with the other main class names used in the project (AdminCLI)
```suggestion
                            <Main-Class>za.co.absa.spline.testdatagen.GenCLI</Main-Class>
```

test-data-generator/src/test/scala/za/co/absa/spline/testdatagen/PlanGeneratorTest.scala
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package za.co.absa.spline.testdatagen
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers._
+import org.scalatest.matchers.should.Matchers.convertToAnyShouldWrapper
+import za.co.absa.spline.producer.model.v1_2.{DataOperation, ExecutionEvent}
+
+class PlanGeneratorTest extends AnyFlatSpec {
```
wajda: By a convention we use `Spec` suffix for naming the Spec-based test suite classes.
```suggestion
class PlanGeneratorSpec extends AnyFlatSpec {
```

test-data-generator/src/test/scala/za/co/absa/spline/testdatagen/PlanGeneratorTest.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  it should "generate read" in {
+    val readOperation = PlanGenerator.generateRead()
+    readOperation.name.getOrElse("") should startWith("generated read")
+    readOperation.output shouldEqual None
+  }
+
+  it should "generate 4 data operations when 4 opCount is provided" in {
+    val id = "23"
+    val dataOperations: Seq[DataOperation] = PlanGenerator.generateDataOperations(4, Seq.empty, Seq(id))
```
wajda: minor: unnecessary variable `id`.
It's only used once and the generic name `id` doesn't add any explanatory context for the reader.
I would recommend using named parameters to make the test code more self-explanatory.
```suggestion
    val dataOperations: Seq[DataOperation] = PlanGenerator.generateDataOperations(opCount = 4, allOps = Seq.empty, childIds = Seq("23"))
```

wajda (approved):
LGTM

PR #1029, opened 2022-02-18
Feature/spline 1000 arango search

### LINKED ISSUE #1000: Consumer API :: Improve search performance
Replace string search with the index search

wajda:
- For text lookups, replace AQL ``` FILTER CONTAINS(...)``` with the ArangoSearch views
- Add `dataSource.lastWriteDetails` property to substitute query time grouping with index time incremental update.
- Move code populating `progress.executionPlanDetails` into a transaction.
- Removed unused persistent indices from the `progress`, `dataSourceCollection` and `operation` collections.
- Updated database migration script accordingly

### DIFF
--- a/commons/src/main/scala/za/co/absa/spline/common/StringEscapeUtils.scala
+++ b/commons/src/main/scala/za/co/absa/spline/common/StringEscapeUtils.scala
@@ -18,6 +18,6 @@ package za.co.absa.spline.common
 
 object StringEscapeUtils {
   def escapeAQLSearch(str: String): String = str
-    .replaceAll("_", "\\\\\\\\_")
-    .replaceAll("%", "\\\\\\\\%")
+    .replace("_", """\\_""")
+    .replace("%", """\\%""")
 }
--- a/commons/src/main/scala/za/co/absa/spline/common/StringEscapeUtils.scala
+++ b/commons/src/main/scala/za/co/absa/spline/common/StringEscapeUtils.scala
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2022 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.common
+
+object StringEscapeUtils {
+  def escapeAQLSearch(str: String): String = str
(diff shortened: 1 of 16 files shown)


commons/src/main/scala/za/co/absa/spline/common/StringEscapeUtils.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.common
+
+object StringEscapeUtils {
+  def escapeAQLSearch(str: String): String = str
+    .replaceAll("_", "\\\\\\\\_")
+    .replaceAll("%", "\\\\\\\\%")
```
cerveada: That is a lot of backslashes. Maybe [Matcher.quoteReplacement](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html#quoteReplacement(java.lang.String)) could be used here as suggested in replaceAll docs. To make it more obvious what is happening.

producer-services/src/main/scala/za/co/absa/spline/producer/service/repo/ExecutionProducerRepositoryImpl.scala
```diff
@@ -186,39 +141,89 @@ object ExecutionProducerRepositoryImpl {
...
+              |      "frameworkName"    : `${ep.systemInfo.name} ${ep.systemInfo.version}`,
+              |      "applicationName"  : ep.name,
+              |      "dataSourceUri"    : dataSourceUri,
+              |      "dataSourceName"   : dataSourceName,
+              |      "dataSourceType"   : dataSourceType,
+              |      "append"           : append
+              |    }
+              |
+              |    if (ep.discriminator != p.discriminator) {
+              |      // nobody should never see this happening, but just in case the universe goes crazy...
```
cerveada:
```suggestion
              |      // nobody should ever see this happening, but just in case the universe goes crazy...
```
Double negative

commons/src/main/scala/za/co/absa/spline/common/StringEscapeUtils.scala
wajda: Not sure if this makes it more obvious :)
```scala
  def escapeAQLSearch(str: String): String = {
    str
      .replaceAll("_", quoteReplacement(quoteReplacement("\\_")))
      .replaceAll("%", quoteReplacement(quoteReplacement("\\%")))
  }
```
What about this then?
```scala
  def escapeAQLSearch(str: String): String = {
    str
      .replaceAll("_", """\\\\_""")
      .replaceAll("%", """\\\\%""")
  }

```

cerveada: What about this:

```scala
  def escapeAQLSearch(str: String): String = {
    str
      .replace("_", """\\_""")
      .replace("%", """\\%""")
  }
```

It seems to be dooing exactly what you want without the need for regexp replacement escaping.

wajda: Lol I must be too enthusiastic about regexps :rofl:

cerveada (approved):

PR #1035, opened 2022-03-07
spline #1034 Consumer API :: Data Source :: multi-value `append` filter

### LINKED ISSUE #1034: Consumer API :: Data Source :: multi-value `append` filter
According to AbsaOSS/spline-ui#179 the filter `append` on a data-source resource should accept any combination of values:
- `true` (append)
- `false` (overwrite)
- `null` (no writes / read-only)

wajda:
Consumer API: Allow for using multi-value filter on the write mode (`append`) property

### DIFF
--- a/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/DataSourcesController.scala
+++ b/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/DataSourcesController.scala
@@ -19,7 +19,7 @@ import io.swagger.annotations.{Api, ApiOperation, ApiParam}
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.web.bind.annotation._
 import za.co.absa.spline.consumer.rest.controller.DataSourcesController.PageableDataSourcesResponse
-import za.co.absa.spline.consumer.rest.model.LabelQuery
+import za.co.absa.spline.consumer.rest.model.{LabelQuery, WriteModeQuery}
 import za.co.absa.spline.consumer.service.model._
 import za.co.absa.spline.consumer.service.repo.DataSourceRepository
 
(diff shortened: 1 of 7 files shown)


consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/DataSourcesController.scala
```diff
@@ -68,8 +68,8 @@ class DataSourcesController @Autowired()(
     @ApiParam(value = "Text to filter the results")
     @RequestParam(value = "searchTerm", required = false) searchTerm: String,
 
-    @ApiParam(value = "Write mode (true - append, false - overwrite")
-    @RequestParam(value = "append", required = false) append: java.lang.Boolean,
+    @ApiParam(value = "Write mode. Comma separated values of: `true` - append, `false` - overwrite, empty string - read only")
+    @RequestParam(value = "append", required = false) append: String,
```
cerveada, line 72: It seems weird that `append = ""` means that it's read. Wouldn't it be better to have three booleans for each or something like enum `APPEND`, `OVERWRITE` and `READ`?

wajda: This API is a subject for a redesign, I don't want to spend much time on it.

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/model/WriteModeQuery.scala
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.spline.consumer.rest.model
+
+object WriteModeQuery {
+  def parse(str: String): Array[Option[Boolean]] = {
+    str.split(",").map(s => {
```
cerveada:
```suggestion
    str.split(",").map(_.trim).map(s => {
```
In rare case, someone sends a space there...

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/DataSourcesController.scala
cerveada: ok

cerveada (dismissed):

consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/model/WriteModeQuery.scala
wajda: That's why one should never save on unit tests :smile: That `trim` used to be there before commit, but somehow vanished at the last moment :smile: :smile: Thank you Adam!

wajda: Ah, no, my bad, it was in another PR

PR #1074, opened 2022-04-04
Feature/spline 1048 admin collection props

### LINKED ISSUE #1048: Admin CLI :: Specify collection props
We need to specify the following collection properties when creating a database:
1. numberOfShards
2. shardKeys
3. replicationFactor
4. waitForSync

wajda:
Add the following options to the `db-init` command of the _Admin CLI_:
```
  --shard-num <value>      Override number of shards per collection. Comma-separated key-value pairs, e.g. 'collectionA=2,collectionB=3'.
  --shard-num-default <value>
                           Override default number of shards.
  --shard-keys <value>     Override shard keys per collection. Comma-separated key-value pairs, where value is a key name list separated by '+', e.g. 'collectionA=k1+k2,collectionB=k3'.
  --shard-keys-default <value>
                           Override default shard keys. Key names separated by '+' character, e.g. 'key1+key2+key3'.
  --repl-factor <value>    Override replication factor per collection. Comma-separated key-value pairs, e.g. 'collectionA=2,collectionB=3'.
  --repl-factor-default <value>
                           Override default replication factor.
  --wait-for-sync          Ensure the data is synchronized to disk before returning from a document CUD operation.

```

resolves #1048

### DIFF
--- a/admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
+++ b/admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
@@ -27,6 +27,7 @@ import za.co.absa.spline.common.scala13.Option
 import za.co.absa.spline.common.security.TLSUtils
 import za.co.absa.spline.persistence.AuxiliaryDBAction._
 import za.co.absa.spline.persistence.OnDBExistsAction.{Drop, Fail, Skip}
+import za.co.absa.spline.persistence.model.CollectionDef
 import za.co.absa.spline.persistence.{ArangoConnectionURL, ArangoManagerFactory, ArangoManagerFactoryImpl}
 
 import scala.annotation.nowarn
@@ -65,6 +66,8 @@ object AdminCLI extends App {
 }
 
 class AdminCLI(dbManagerFactory: ArangoManagerFactory) {
+
+  import za.co.absa.spline.common.CollectionUtils.Implicits._
 
   def exec(args: Array[String]): Unit = {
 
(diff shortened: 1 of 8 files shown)


admin/src/main/scala/za/co/absa/spline/admin/AdminCLI.scala
```diff
@@ -107,7 +110,65 @@ class AdminCLI(dbManagerFactory: ArangoManagerFactory) {
...
+        opt[Map[String, Int]]("shard-num")
+          text "Override number of shards per collection. Comma-separated key-value pairs, e.g. 'collectionA=2,collectionB=3'."
+          validate (
+          _.values.collectFirst({
+            case v if v < 1 => failure(s"Shard number should be positive, but was $v")
+          }).getOrElse(success))
+          action {
+          case (m, c@AdminCLIConfig(cmd@DBInit(_, _, _, opts), _, _)) =>
+            c.copy(cmd.copy(options = opts.copy(numShards = m.mapKeys(CollectionDef.forName))))
+        },
```
cerveada: The indentation seems wrong. 

```suggestion
        opt[Map[String, Int]]("shard-num")
          text "Override number of shards per collection. Comma-separated key-value pairs, e.g. 'collectionA=2,collectionB=3'."
          validate (
            _.values.collectFirst({
              case v if v < 1 => failure(s"Shard number should be positive, but was $v")
            }).getOrElse(success))
          action {
            case (m, c@AdminCLIConfig(cmd@DBInit(_, _, _, opts), _, _)) =>
              c.copy(cmd.copy(options = opts.copy(numShards = m.mapKeys(CollectionDef.forName))))
          },
```

wajda: It's a long standing IDEA auto-layout problem. I tried different settings, but it didn't seem it can do it properly for infix notation in Scala code. So I've switched to prefix notation and layout is now correct.

cerveada (approved):

Issue #540, opened 2020-01-02
Data Impact / Forward lineage analysis


wajda:
Opposite and complimentary to the (Backward) Data Lineage.
Traces how the data is used / errors are propagated down the pipeline

wajda:
Probably. The work on 0.6 has been suspended for the last month or so due to performance issues found in 0.5.3 that affect some production usages and had to be addressed. We'll have another round of 0.6 planning somewhere next week and clarify on this issue.

pratapmmmec:
Thanks for confirming @wajda . Fingers crossed for this epic.

dk1844:
The backend part of forward lineage (aka impact) on the service-side merged to `develop`. Note, that there is no UI support for this feature, yet.

PR #1091, opened 2022-07-14
Feature/540 forward lineage

dk1844:
Heavily inspired by (backward) Lineage. After the naive implementation, common parts of the foxx service have been moved to `commons.js` for reuse.

(Also, there is a similar PR #1092 - let's inspire each other, @vishalag001)

### Spline service level API
`GET /spline/events/$eventId/lineage-overview/$maxDepth`
 - yields the impact (forward lineage) graph object
 
### ArangoDB foxx-service level API
`GET /events/:eventKey/impact-overview/:maxDepth`
 - yields the impact (forward lineage) graph object (from `persistence`)

## Test-run
I have hacked the spline-ui to load `.../impact-overview` url instead of `.../lineage-overview` to test the graph data in UI to be shown:

hacked-ui-to-show-impact-overview

Implements backed of #540.


dk1844:
@wajda @cerveada At this point, the solution seems to work. Is it worth, or rather, to what degree should I try to deduplicate with the original code - there are quite obviously some similar parts...?

wajda:
> @wajda @cerveada At this point, the solution seems to work. Is it worth, or rather, to what degree should I try to deduplicate with the original code - there are quite obviously some similar parts...?

It's up to you. Of course the code quality is important, and I don't think any AQL or FOXX code has any special additions or exceptions related to the common code quality standards. Just use a usual common sense. When you're done mark it as "ready for review" and we'll take a look.

cerveada:
What is a purpose of the two empty files?

dk1844:
> What is a purpose of the two empty files?

After some digging, I have discovered that it was due to file mode being inadvertently changed - probably when sharing between Win-Linux/etc while having git's `core.filemode` enabled. Some details [here](https://stackoverflow.com/a/56388238). The default settings always changed my file mode even if I reverted any changes.

So I have set `git config core.filemode false` on my cloned repo on Windows and I have reverted the files to the original state 👍 . Thanks for noticing

cerveada (approved):

dk1844:
Closed in favor of https://github.com/AbsaOSS/spline/pull/1100

PR #1092, opened 2022-07-18
Forward Lineage

vishalag001:
Following up on the discussion: https://github.com/AbsaOSS/spline/discussions/1081
I created a new foxx service which gives the forward lineage for a given write event. I was able to visualise the forward lineage on the lineage overview page.

Attaching the screenshot of the same.

forwardLineage

The PR is for the backend change for the forward lineage.

### DIFF
--- a/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageOverviewController.scala
+++ b/consumer-rest-core/src/main/scala/za/co/absa/spline/consumer/rest/controller/LineageOverviewController.scala
@@ -53,4 +53,28 @@ class LineageOverviewController @Autowired()(val repo: LineageRepository) {
     maxDepth: Int
   ): Future[LineageOverview] =
     repo.lineageOverviewForExecutionEvent(eventId, maxDepth)
+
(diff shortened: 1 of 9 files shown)


dk1844:
Great work, @vishalag001, thank you for providing this!

I have actually been working on something similar and few days ago, I have started pushing this, too - in https://github.com/AbsaOSS/spline/pull/1091. 

I am sure, that we can learn from each other's codebases and finish the feature successfully, now, here or there! 🤝

dk1844:
So have actually went through this PR and it is *extremely* similar to #1091 (mostly there is just difference in wording or file naming).

There is one important difference, though, it is how [observedReadsByWrite](https://github.com/AbsaOSS/spline/blob/cc4a60a0f0dba8d286f39b5af355a64f065cdf40/persistence/src/main/resources/foxx/spline/services/observed-reads-by-write.js) are defined - not only we need to assess reads happening after the initial write, but also, we need to disregard read after a rewrite (since the lineage would get broken at that point)

@vishalag001, would you mind reviewing #1091, too?

wajda:
I think we can close it in favour of #1091 if you guys don't mind.

PR #1099, opened 2022-07-21
Feature/1098 foxx module

### LINKED ISSUE #1098: Move AQL functions and Foxx services to separate modules
Create separate TypeScript based modules for Foxx and AQL functions.

wajda:
fixes #1098

- move Foxx and AQL functions to separate modules
- migrate from JavaScript to TypeScript codebase
- setup build/test/lint phases for the projects
- unify nodejs version used for building various modules in the project


admin/src/main/scala/za/co/absa/spline/arango/foxx/FoxxManagerImpl.scala
```diff
@@ -31,23 +27,10 @@ class FoxxManagerImpl(restClient: RESTClient)(implicit ec: ExecutionContext)
   extends FoxxManager
     with Logging {
 
-  override def install(mountPrefix: String, assets: Array[(String, String)]): Future[Unit] = {
+  override def install(mountPrefix: String, content: Array[Byte]): Future[Unit] = {
```
dk1844, line 30: It took me a while to realize why the zipping is not needed anymore (using `maven-assembly-plugin` with `<format>zip</format>` defined in new modules' `assembly.xml`).

Is there an obvious upside for this approach? I know, it is declarative, but it seems that the configuration to achieve this is quite big.

arangodb-foxx-services/src/main/services/lineage-overview.ts
```diff
@@ -28,55 +30,55 @@ const {observedWritesByRead} = require('./observed-writes-by-read');
  * It shows how far the traversal should looks for the lineage.
  * @returns za.co.absa.spline.consumer.service.model.LineageOverview
  */
-function lineageOverview(eventKey, maxDepth) {
+export function lineageOverview(eventKey: DocumentKey, maxDepth: number) {
```
dk1844, line 33: With  typescript's `export` markup, the following at the end of the file should not be needed, I assume:
```javascript
module.exports = {
    lineageOverview
}
```

dk1844 (dismissed):
LGTM (just read the code)

I like change very much, especially employing TypeScript! I am not even mad I might need redo the https://github.com/AbsaOSS/spline/pull/1091.

wajda:
> LGTM (just read the code)
> 
> I like change very much, especially employing TypeScript! I am not even mad I might need redo the #1091.

lol, not redoing, mostly just moving and changing file extensions :D

admin/src/main/scala/za/co/absa/spline/arango/foxx/FoxxManagerImpl.scala
wajda: Well, it was not done on purpose, I just made use of the fact the module is now packed as zip anyway. The intention was to split a TS based Foxx code from Scala based application level code running on a different tier. And since TS is not Java, the most logical distribution format is ZIP. And the Foxx package format is ZIP. So why would we pack-unpack-pack then?

dk1844: 😆 oh, I see

arangodb-foxx-services/src/main/services/lineage-overview.ts
wajda: Absolutely! That's a leftover, thanks for spotting it.

arangodb-aql-functions/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+
+    <parent>
+        <groupId>za.co.absa.spline</groupId>
+        <artifactId>parent-pom</artifactId>
+        <relativePath>../build/parent-pom/pom.xml</relativePath>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>arangodb-aql-functions</artifactId>
+    <packaging>pom</packaging>
```
cerveada, line 31: Shouldn't the packaging be zip?

wajda:
I also added `foxx-cli` scripts for development convenience, to quickly build and deploy the Foxx service with one short command without involving full-fledged Maven build and Admin CLI.

```shell
npm foxx:deploy --database blah
```

arangodb-aql-functions/pom.xml
wajda: no-no, that's how it should be. "Zip" is a type of the artifact, but "pom" is a type of the module.

cerveada (approved):

PR #1100, opened 2022-07-26
Feature/540 forward lineage ts + cross-env

dk1844:
Heavily inspired by (backward) Lineage. After the naive implementation, common parts of the foxx service have been moved to `commons.ts` for reuse.

(Also, there is a similar PR #1092 - let's inspire each other, @vishalag001)

### Spline service level API
`GET /spline/events/$eventId/lineage-overview/$maxDepth`
 - yields the impact (forward lineage) graph object
 
### ArangoDB foxx-service level API
`GET /events/:eventKey/impact-overview/:maxDepth`
 - yields the impact (forward lineage) graph object (from `persistence`)

## Test-run
I have hacked the spline-ui to load `.../impact-overview` url instead of `.../lineage-overview` to test the graph data in UI to be shown:

hacked-ui-to-show-impact-overview

## Cross-env usage:
NPM package [cross-env](https://www.npmjs.com/package/cross-env) was also added for `rest-service` and `arangodb-foxx-services` modules, because the npm original run stripts were targetting POSIX environments only; now the build runs correctly on Windows, too, specifically:
  - swagger docs are now generated properly on Windows
  - as dev-support, you are able to run `npm run foxx:deploy --database=NAME` in `arangodb-foxx-services` directory to reinstall foxx services.

Implements backed of #540.

### DIFF
--- a/arangodb-foxx-services/src/main/model.ts
+++ b/arangodb-foxx-services/src/main/model.ts
@@ -15,45 +15,49 @@
  */
 
 export type DocumentKey = string
-export type ExecutionEvent = ExecutionEventLike
-export type DataSource = DataSourceLike
-
-interface DataSourceLike {
-    _key: string
-}
-
-interface ExecutionEventExtra {
-    appId: string
-}
-
-export interface ExecutionEventLike {
+export type ExecutionEvent = {
     timestamp: bigint
-    extra: ExecutionEventExtra
+    extra: {
+        appId: string
+    }
+}
+export type DataSource = {
+    _key: string
+    name: string
+}
+
+export type ExecutionNode = {
+    _key: string
+    name: string
+
+}
+export type LineageGraphNode = ExecutionNode | DataSource
+
+export type LineageGraphEdge = {
+    source: string
+    target: string
+}
+
+export type LineageGraph = {
+    depth: number
+    vertices: Array<LineageGraphNode>
+    edges: Array<LineageGraphEdge>
 }
 
 /**
- * mimics backend's [[za.co.absa.spline.consumer.service.model.LineageOverview]]
+ * Mimics backend's [[za.co.absa.spline.consumer.service.model.LineageOverview]]
  */
-export class LineageOverview {
-    constructor(
-        readonly info: LineageOverviewInfo,
-        readonly graph: LineageOverviewGraph) {
+export type LineageOverview = {
+    info: {
+        timestamp: bigint
+        applicationId: string
+        targetDataSourceId: string
+    }
+    graph: {
+        depthRequested: number
+        depthComputed: number
+        nodes: Array<LineageGraphNode>
+        edges: Array<LineageGraphEdge>
     }
 }
 
-export class LineageOverviewInfo {
-    constructor(
-        readonly timestamp: bigint,
-        readonly applicationId: string,
-        readonly targetDataSourceId: string) {
-    }
-}
-
-export class LineageOverviewGraph {
-    constructor(
-        readonly depthRequested: number,
-        readonly depthComputed: number,
-        readonly nodes: Array<object>,
-        readonly edges: Array<object>) {
-    }
-}
--- a/arangodb-foxx-services/src/main/model.ts
+++ b/arangodb-foxx-services/src/main/model.ts
@@ -35,37 +35,25 @@ export interface ExecutionEventLike {
  * mimics backend's [[za.co.absa.spline.consumer.service.model.LineageOverview]]
  */
 export class LineageOverview {
-    readonly info: LineageOverviewInfo
-    readonly graph: LineageOverviewGraph
-
-    constructor(info: LineageOverviewInfo, graph: LineageOverviewGraph) {
-        this.info = info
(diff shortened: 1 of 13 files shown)


arangodb-foxx-services/src/main/model.ts
```diff
@@ -15,4 +15,57 @@
...
+    readonly nodes: Array<object>
+    readonly edges: Array<object>
+
+    constructor(depthRequested: number, depthComputed: number, nodes: Array<object>, edges: Array<object>) {
+        this.depthRequested = depthRequested
+        this.depthComputed = depthComputed
+        this.nodes = nodes
+        this.edges = edges
+    }
+}
```
wajda: minor suggestion: 
You can write the same definition with less lines of code.
```suggestion
export class LineageOverviewGraph {
    constructor(
        readonly depthRequested: number,
        readonly depthComputed: number,
        readonly nodes: Array<object>,
        readonly edges: Array<object>) {
    }
}
```

wajda: In general, why do you use classes/interfaces for merely defining complex data types? Classes and interfaces are OO abstractions and are useful when you add behavior (especially interfaces are there to describe public contract primarily in terms of methods). In this case, I would rather use just `type`

arangodb-foxx-services/src/main/model.ts
```diff
@@ -15,4 +15,57 @@
...
+        this.applicationId = applicationId
+        this.targetDataSourceId = targetDataSourceId
+    }
+}
+
+export class LineageOverviewGraph {
+    readonly depthRequested: number
+    readonly depthComputed: number
+    readonly nodes: Array<object>
+    readonly edges: Array<object>
```
wajda: Can we use algebraic data types instead of those objects?

arangodb-foxx-services/src/main/services/observed-reads-by-write.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Returns a list of execution events which reads are visible from the write of the given execution event
+ *
+ * @param writeEvent za.co.absa.spline.persistence.model.Progress
+ * @returns za.co.absa.spline.persistence.model.Progress[]
+ */
+export function observedReadsByWrite(writeEvent) {
+    return writeEvent && db._query(aql`
+        FOR wExPlan IN 1 OUTBOUND ${writeEvent} progressOf
```
wajda, line 29: Missing `WITH` statement

wajda: nit: capitalization

arangodb-foxx-services/src/main/model.ts
dk1844: The suggested change - amazing, I didn't know this - I tried to learn TS basics from https://www.typescriptlang.org/docs/handbook/2/classes.html, but that was not apparent.

I use the classes/interfaces because the object I access fields on, e.g. `executionEvent.timestamp` -- and the linter/IDE warns me about that field not being present - so I define a (at least partially) the class/interface to have the fields as part of the contract.

```typescript
export type ExecutionEvent = object
executionEvent.timestamp
```
gets you a warning (not expecting for `object.timestamp` to be available/correct), while
```typescript
export type ExecutionEvent = ExecutionEventLike
export interface ExecutionEventLike {
    timestamp: bigint
    ...
}
executionEvent.timestamp
```
seems to be ok (the contract is defined by the interface).

Or maybe I am missing something and you mean to use `type` differently?

arangodb-foxx-services/src/main/services/observed-reads-by-write.ts
dk1844: Thanks, both fixed.
Surprisingly, it worked without `WITH` 😮

arangodb-foxx-services/src/main/model.ts
dk1844: Suggestion applied in all similar constuctors, thanks

arangodb-foxx-services/src/main/model.ts
dk1844: Thanks for the tip and explanation, done.

arangodb-foxx-services/src/main/model.ts
dk1844: Thanks for the tip, I have used types as suggested in the end.

cerveada (dismissed):

arangodb-foxx-services/src/main/services/observed-reads-by-write.ts
wajda: By default, it's automatically inferred on the single server deployments, but is required on clusters.
So when testing on a local ArangoDB set the `query.require-with=true` to force it to behave more like it will on a real prod setup.
E.g. in the `arangod.conf` under the "query" section:
```properties
[query]
require-with = true
```

cerveada (dismissed):

wajda (dismissed):
LGTM

wajda (approved):

PR #1047, opened 2022-03-23
Feature/spline #1017 docker generator

AdrianOlosutean:
fixes #1017

PR #1102, opened 2022-08-29
Feature/1017 data generator output name

dk1844:
This is an update of #1047, it introduces a few changes:
 - range values are not expressed as `FROM-TO/STEP`, but `FROM-TO@STEP`. This prevented being used on FSs where files cannot contain `/`.
 - new option `-t`/`--output` is introduced in order for the user to be able to choose their own output filename in a controlled manned and would not have to guess or derive what the result will be.

### Sidenote
My ultimate aim is to merge this, later with #1047, too. This will allow having a docker image of the data-generator later usable in related project [Spline Getting Started](https://github.com/AbsaOSS/spline-getting-started/) for performance measuring.


test-data-generator/src/main/scala/za/co/absa/spline/testdatagen/GenCLI.scala
```diff
@@ -101,9 +105,14 @@ object GenCLI {
...
-        s"${config.operations}ops-" +
-        s"${config.attributes}attr")
+      config.customOutputFileName.fold {
+        val defaultPrefixName = s"${config.graphType.name}-lineage-" +
+          s"${config.reads}reads-" +
+          s"${config.operations}ops-" +
+          s"${config.attributes}attr"
+        new FileDispatcher(defaultPrefixName) // using default prefix and default built-in suffix
+      } { customFileName =>
+        new FileDispatcher(customFileName, "") // custom name as-is
```
cerveada: the filename creation seems to be somehow spread over two classes. I think better would be to just have a method for creating the default filename (including the suffix).
```Scala
new FileDispatcher(createDefaultFileName(config))
```

Then you don't need `fileNameSuffix` in FileDispatcher

dk1844: 👍 Good idea, done.

cerveada (approved):

PR #1105, opened 2022-09-20
curl added to rest-gateway dockerfile

dk1844:
This PR installs `curl` inside Spline's `rest-gateway` docker image in order to correctly use it during the docker-compose orchestration.

At least on my machine (Win 10 box), curl was not present and the orchestration has been failing due to this fact

### DIFF
--- a/rest-gateway/Dockerfile
+++ b/rest-gateway/Dockerfile
@@ -18,6 +18,8 @@ ARG DOCKER_BASE_IMAGE_PREFIX=
 
 FROM "$DOCKER_BASE_IMAGE_PREFIX"tomcat:9-jre11-openjdk-slim-buster
 
+RUN apt-get update && apt-get install curl -y
+
 LABEL \
     vendor="ABSA" \
     copyright="2019 ABSA Group Limited" \


wajda:
As discussed let's find another way of doing healthcheck that doesn't involve installing `curl` command. This topic seems to be currently discussed on the internet.
One of the alternative method that I've found is this:
```bash
timeout 10s bash -c ':> /dev/tcp/127.0.0.1/8080'
```
Ref: https://medium.com/bash-tips-and-tricks/part01-tcp-udp-request-with-a-native-bash-feature-and-without-curl-wget-9dcef59c30aa

dk1844:
> As discussed let's find another way of doing healthcheck that doesn't involve installing `curl` command. This topic seems to be currently discussed on the internet. One of the alternative method that I've found is this:
> 
> ```shell
> timeout 10s bash -c ':> /dev/tcp/127.0.0.1/8080'
> ```
> 
> Ref: https://medium.com/bash-tips-and-tricks/part01-tcp-udp-request-with-a-native-bash-feature-and-without-curl-wget-9dcef59c30aa

This is very interesting, I will try it out. Thanks 👍

dk1844:
Closing in favor of using a different solution for the testing script (custom `Dockerfile` for testing vs using bash-based health checks and docker-compose `depends_on`)

PR #1107, opened 2022-09-28
Feature/spline 954 acid tx

### LINKED ISSUE #954: Optimize transaction size + ACID on cluster
For performance and scaling reasons it's important to keep ArangoDB transaction small. Currently we insert the entire execution plan graph in a single transaction which sometimes may take longer than necessary depending of the number and size of nodes being inserted.
To reduce the transaction size to the minimum we could insert majority of documents non-transactionally, and close the action by inserting the remaining `ExecutionPlan` and edges to the `DataSource` in a small fast transaction.

wajda:
Fixes #954 

1. Removed `arangodb-AQL-functions` module and moved the code to the `arangodb-foxx-services`
2. Moved custom AQL functions lifecycle management responsibility from the _Admin CLI_ to the Foxx setup/teardown scripts.
3. Moved lineage ingestion functionality from the _Producer_ service layer to the Foxx layer
4. Changed the lineage persistence logic according to the proposed application level ACID Tx protocol.


arangodb-foxx-api/scala-ts.conf
```diff
@@ -0,0 +1,17 @@
+compilationRuleSet {
+   includes = [ ".*\\.scala", "Transpiler.*" ]
+   excludes = [ "foo" ]
```
cerveada: Is the `foo` supposed to be there?

arangodb-foxx-api/src/main/scala/za/co/absa/spline/persistence/model/Attribute.scala
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 ABSA Group Limited
+ * Copyright 2022 ABSA Group Limited
```
cerveada: Should the year change when the file is moved?

arangodb-foxx-services/src/main/services/commons.ts
```diff
@@ -13,29 +13,30 @@
...
 import {aql, db} from '@arangodb'
 import {memoize} from '../utils/common'
 import {GraphBuilder} from '../utils/graph'
-import {AnyFunction} from '../utils/types'
+import { ReadTxInfo } from '../persistence/model'
+import { Progress } from '../../external/api.model'
 
 
-export function getExecutionEventFromEventKey(eventKey: DocumentKey): ExecutionEvent {
+export function getExecutionEventFromEventKey(eventKey: DocumentKey): Progress {
```
cerveada, line 25: What is the reason behind this change `ExecutionEvent  -> Progress` ?

arangodb-foxx-services/src/scripts/aql-functions.d.ts
```diff
@@ -0,0 +1,56 @@
+/*
```
cerveada, line 1: What is the meaning of the `d` int the name?

persistence/src/test/scala/za/co/absa/spline/persistence/tx/JSTxBuilderSpec.scala
```diff
@@ -19,40 +19,41 @@ package za.co.absa.spline.persistence.tx
...
   it should "generate NATIVE statements" in {
-    val generatedJS = new TxBuilder()
-      .addQuery(NativeQuery("db.FOO();"))
+    val generatedJS = new JSTxBuilder()
+      .addQuery(NativeQuery("return db.FOO();"))
       .addQuery(NativeQuery("db.BAR();"))
       .generateJs()
 
-    generatedJS should be {
+    generatedJS shouldEqual {
```
cerveada, line 34: What is the reason for this change?

arangodb-foxx-api/scala-ts.conf
wajda: The whole file is a leftover from experimentation with another `scala-ts` transpiler, and is not needed.

arangodb-foxx-services/src/main/services/commons.ts
wajda: Just eliminating duplicated type definition. More changes are yet to be made to the model as different parts of it have being moved from other modules and consolidated on the Foxx layer.

arangodb-foxx-services/src/scripts/aql-functions.d.ts
wajda: I was struggling to find an understandable name for the directory and decided to just name it after the type of the files it contains. It's temporary here, I'm going to remove it as soon as it's promoted to the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/arangodb) project.

`.d.ts` - is a standard _TypeScript_ type declaration file extension. In this case it's extra definition of the Foxx API that are missing form the latest version of the `@types/arangodb` package

persistence/src/test/scala/za/co/absa/spline/persistence/tx/JSTxBuilderSpec.scala
wajda: consistency. The other spec statements in the file use `shouldEqual`. Moreover to me it seems more appropriate and for the string comparison assertion, while `be` implies a predicate following it.

cerveada: There seems to be a functional difference between `equal` and `be`.
AS I understand it `be` always works the same as `==` in Scala, whereas `equal` will use a user-defined implicit `Equality` if available in scope. 

Maybe it would be better to use `be` as a default and `equal` only when we need the additional functionality?

https://www.scalatest.org/user_guide/using_matchers#checkingEqualityWithMatchers

arangodb-foxx-services/src/main/services/observed-reads-by-write.ts
```diff
@@ -15,47 +15,64 @@
...
+                FOR readOp IN operation
+                    ${aqlGen.genTxIsolationCodeForLoop('readOp')}
+                    FILTER readOp.type == 'Read'
+                    FILTER readOp._belongsTo == planId
+                    FOR ds IN 1 OUTBOUND readOp readsFrom
+                        FILTER ds._key == wds._key
+                        RETURN DISTINCT readEvent
+    `
+
+    Logger.debug(query)
```
korel-san, line 75: @wajda debug

arangodb-foxx-services/src/main/services/observed-writes-by-read.ts
```diff
@@ -15,40 +15,54 @@
...
+                        SORT p.timestamp ASC
+                        RETURN p
                 )
             LET allObservedEvents = APPEND(maybeObservedOverwrite, observedAppends)
-            FOR e IN allObservedEvents RETURN e
-    `).toArray()
+            FOR p IN allObservedEvents RETURN p
+    `
+
+    Logger.debug(query)
```
korel-san, line 65: @wajda do you really want to leave it in the codebase?

arangodb-foxx-services/src/scripts/fs.d.ts
```diff
@@ -0,0 +1,337 @@
+/*
```
korel-san, line 1: @wajda is it typings for the fs module?

korel-san: if yes, I would recommend leaving @types/node in package.json and in the tsconfig.json
https://stackoverflow.com/questions/43048113/use-fs-in-typescript

persistence/src/main/scala/za/co/absa/spline/persistence/tx/AbstractTxBuilder.scala
```diff
@@ -0,0 +1,49 @@
+/*
```
korel-san, line 1: @wajda I would recommend reorganizing the number of subfolders, due to the complexity and avoiding the Doorway Effect

korel-san (approved):

dk1844 (approved):
Read the code, have not found anything suspicious. But I must admit that I do not understand every bit of the code.

arangodb-foxx-services/src/main/services/observed-reads-by-write.ts
wajda: ... so?

arangodb-foxx-services/src/main/services/observed-writes-by-read.ts
wajda: Yes, why?

arangodb-foxx-services/src/scripts/fs.d.ts
wajda: I tried that but unfortunately it didn't work. ArangoDB for some reason uses a slightly different [File system API](https://www.arangodb.com/docs/stable/appendix-java-script-modules-file-system.html).

> The implementation tries to follow the CommonJS [Filesystem/A/0](http://wiki.commonjs.org/wiki/Filesystem/A/0) specification where possible.

Not sure how many _specs_ for the FS module exist in the JavaScript world, but the one `@types/node` provides doesn't fit what ArangoDB is using. My intention for this type-def file was to replace it with a library one as soon as possible (either I find something existing, or make a PR to the `@types/arangodb`)

persistence/src/test/scala/za/co/absa/spline/persistence/tx/JSTxBuilderSpec.scala
wajda: We could. I don't see too much practical difference to be honest, but I'm fine with either. I more care about consistency though, so that the same logic of choice is applied throughout the whole code base. We can discuss it on CQC.

persistence/src/main/scala/za/co/absa/spline/persistence/tx/AbstractTxBuilder.scala
wajda: Sure! Would you suggest anything specific? :)

PR #1113, opened 2022-10-18
Feature/spline 684 data retention 2

### LINKED ISSUE #684: Data retention CLI
Is there any easy way to configure automatic retention for spline data? For example automatically delete entires older than 3 months ?

dk1844:
This PR brings back unmerged PR https://github.com/AbsaOSS/spline/pull/762.

The original service wrapper has been generally used (while brought up to date with latest develop). The internals of the foxx service where the db-prune actually happens has been taken from https://gist.github.com/Aditya-Sood/ecc07c9f296dbdf03d4946c5d1b4efce (https://github.com/AbsaOSS/spline/issues/684#issuecomment-1253274502).

Naively tested.

Based on the [measuring of the implementation options for Stage 1 and Stage 2](https://github.com/AbsaOSS/spline/issues/684#issuecomment-1311499371), _Option A for Stage 1_ and _Option A for Stage 2_ have been selected for the implementation.


arangodb-foxx-services/src/main/routes/admin-router.ts
```diff
@@ -0,0 +1,35 @@
...
+adminRouter
+    .delete('/data/before/:timestamp',
+        (req: Foxx.Request, res: Foxx.Response) => {
+            const timestamp = req.pathParams.timestamp
+            pruneBefore(timestamp)
+            res.send('DB Pruning request submitted.')
+        })
+    .pathParam('timestamp', joi.number().integer().min(0).required(), 'Data retention threshold [timestamp in millis]')
+    .summary('Prune database')
+    .description('Garbage collect the data older than the given timestamp')
```
wajda: Minor (wording): to me _garbage collection_ means reclaiming resources occupied by unused and unreferenced data. however in our case, we remove expired, but still referenced and hence potentially used stuff. So I would rather call it just that - "Delete", not "garbage collect"

arangodb-foxx-services/src/main/routes/admin-router.ts
```diff
@@ -0,0 +1,35 @@
...
+
+
+const adminRouter = createRouter()
+
+adminRouter
+    .delete('/data/before/:timestamp',
+        (req: Foxx.Request, res: Foxx.Response) => {
+            const timestamp = req.pathParams.timestamp
+            pruneBefore(timestamp)
+            res.send('DB Pruning request submitted.')
```
wajda: Minor (wording): Why "request submitted"? The call is synchronous, so I guess the response should be 
"Pruning done".

admin/src/main/scala/za/co/absa/spline/arango/DataRetentionManager.scala
```diff
@@ -0,0 +1,31 @@
...
+
+import com.arangodb.async.ArangoDatabaseAsync
+import com.arangodb.internal.InternalArangoDatabaseOps
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class DataRetentionManager(db: ArangoDatabaseAsync)(implicit ec: ExecutionContext) {
+
+  def pruneBefore(timestamp: Long): Future[Unit] = {
+    new InternalArangoDatabaseOps(db).restClient.delete(s"spline/admin/data/before/$timestamp")
```
wajda: `InternalArangoDatabaseOps` is an implicit wrapper
```suggestion
    import za.co.absa.spline.persistence.ArangoImplicits._
    db.restClient.delete(s"spline/admin/data/before/$timestamp")
```

arangodb-foxx-services/src/main/services/prune-database.ts
```diff
@@ -0,0 +1,130 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {aql, db} from '@arangodb'
+
+export function pruneBefore(timestamp) {
```
wajda, line 20: Was there any specific reason why you preferred breaking one complex AQL query into a series of smaller queries?
Although such approach is probably better from the ArangoDB memory perspective, but I am worried about transferring all those intermediate results (IDs) from AQL engine to V8 and back, which I'm sure are not sharing any memory. Also, it could result in less optimal AQL execution plan as the AQL optimizer does not see into the Foxx function.

I would suggest to try to combine some queries into bigger blocks.

arangodb-foxx-services/src/main/services/prune-database.ts
```diff
@@ -0,0 +1,130 @@
...
+                REMOVE po IN progressOf
+                RETURN DISTINCT po._to
+    `).toArray()
+
+    const t1 = Date.now()
+    console.log(`Purged ${refPlanIds.length} plans, ${t1 - t0} ms`)
+
+    // STAGE 2: No progressOf corresponding to executionPlan (Need to scan all plans, and progressOf foreach) -> delete executionPlan and related fields
+
+    const execPlanIDsWithProgressEventsArray = db._query(`
```
wajda: Missing `aql`?
```suggestion
    const execPlanIDsWithProgressEventsArray = db._query(aql`
```

arangodb-foxx-services/src/main/services/prune-database.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    const dataSourceKeysInAffectsDependsUseArray = dataSourceKeysInAffectUseArray.concat(
+        db._query(`
+            FOR source IN dataSource
+                FOR dependEdge in depends
+                    FILTER dependEdge._to == source._id
+                    COLLECT keys = source._key
+                    RETURN keys
+        `).toArray()
+    )
```
wajda: These two queries are independent, so if there were placed into the same AQL query it potentially could be executed in parallel (in a paid version I guess, but still). In Foxx, however, all calls are strictly synchronous.

arangodb-foxx-services/src/main/services/prune-database.ts
```diff
@@ -0,0 +1,130 @@
...
+    const dataSourceKeysInAffectUseArray = db._query(`
+        FOR source IN dataSource
+            FOR affectEdge in affects
+                FILTER affectEdge._to == source._id
+                COLLECT keys = source._key
+                RETURN keys
+    `).toArray()
+
+    const dataSourceKeysInAffectsDependsUseArray = dataSourceKeysInAffectUseArray.concat(
+        db._query(`
```
wajda: Minor: using a back tick for string literal, should be just `'`

arangodb-foxx-services/src/main/services/prune-database.ts
```diff
@@ -0,0 +1,130 @@
...
+    // Deleted other collections and then the orphanExecPlanIDs
+
+    const t2 = Date.now()
+    console.log(`All collections but dataSource purge in ${t2 - t1} ms`)
+
+    ////////////////////////////////////////////////////////////////////////
+
+
+    // STAGE 3: No `affects` and `depends` corresponding to dataSource (Need to scan all dataSources, and both edges foreach) -> delete dataSource
+    const dataSourceKeysInAffectUseArray = db._query(`
```
wajda: Minor: using a back tick for string literal, should be just `'`

arangodb-foxx-services/src/main/services/prune-database.ts
dk1844: Thanks for noticing, added here and elsewhere.

admin/src/main/scala/za/co/absa/spline/arango/DataRetentionManager.scala
dk1844: Thanks, suggestion applied (although, I have place the import at the top scoped)

arangodb-foxx-services/src/main/routes/admin-router.ts
dk1844: Thanks, suggestion applied.

arangodb-foxx-services/src/main/routes/admin-router.ts
dk1844: Thanks, suggestion applied.

arangodb-foxx-services/src/main/services/prune-database.ts
dk1844: In order to avoid squashing everything to one line or concatenating multiple lines of literal strings, backticks or `aql`backticks seem appropriate -- looking on the doc https://www.arangodb.com/docs/stable/foxx-guides-queries.html

arangodb-foxx-services/src/main/services/prune-database.ts
dk1844: dtto

arangodb-foxx-services/src/main/services/prune-database.ts
wajda: Ah, my bad, sorry. I somehow overlooked it was a multiline AQL query.

arangodb-foxx-services/src/main/services/prune-database.ts
dk1844: Done.

arangodb-foxx-services/src/main/services/prune-database.ts
dk1844: This is maily because the most of the code comes from https://gist.github.com/Aditya-Sood/ecc07c9f296dbdf03d4946c5d1b4efce script - adapted only where needed for our needs.

dk1844: I thought about it some more, for example in case of the looping for the collections to purge in stage 2, that can be done in AQL -- but in that case we would have to forgo the logging (I don't know a way to write to logs from AQL directly).

arangodb-foxx-services/src/main/services/prune-database.ts
```diff
@@ -0,0 +1,131 @@
...
+            FILTER p.timestamp < ${timestamp}
+            REMOVE p IN progress
+            FOR po IN progressOf
+                FILTER po._from == p._id
+                REMOVE po IN progressOf
+                RETURN DISTINCT po._to
+    `).toArray()
+
+    const t1 = Date.now()
+    console.log(`Purged ${refPlanIds.length} plans, ${t1 - t0} ms`)
```
wajda: Let's use `Logger` helper object instead of `console` for logging.

```typescript
    import * as Logger from '../utils/logger'
    ...
    Logger.info(`Purged ${refPlanIds.length} plans, ${t1 - t0} ms`)
```

wajda (commented):
LGTM, excepts from a minor logging comment

arangodb-foxx-services/src/main/services/prune-database.ts
dk1844: Thanks, redone as suggested. I have test-ran it with this logging, too, and witnessed it correctly shows log entries on the ArangoDB LOG UI.

admin/src/main/scala/za/co/absa/spline/admin/DateTimeUtils.scala
```diff
@@ -0,0 +1,60 @@
...
+      val ZonedDateTimeRegexp(ldt, tzOffset, tzId) = s
+      val maybeTzIds = Seq(tzId, tzOffset).map(Option.apply)
+
+      require(!maybeTzIds.forall(_.isDefined), "Either timezone ID or offset should be specified, not both")
+
+      val tz = maybeTzIds
+        .collectFirst({ case Some(v) => ZoneId.of(v) })
+        .getOrElse(defaultZoneId)
+
+      ZonedDateTime.parse(ldt, ZonedDateTimeFormatter.withZone(tz))
```
cerveada, line 55: Is the formatter necessary? It seem like it is the same as default behaviour anyway or did I miss something?

admin/src/main/scala/za/co/absa/spline/admin/commands.scala
```diff
@@ -67,3 +70,12 @@ case class DBExec(
 
   def addAction(action: AuxiliaryDBAction): DBExec = copy(actions = actions :+ action)
 }
+
+case class DBPrune(
+                    override val dbUrl: Url = null,
+                    retentionPeriod: Option[Duration] = None,
+                    thresholdDate: Option[ZonedDateTime] = None
+                  ) extends DBCommand {
```
cerveada:
```suggestion
case class DBPrune(
  override val dbUrl: Url = null,
  retentionPeriod: Option[Duration] = None,
  thresholdDate: Option[ZonedDateTime] = None
) extends DBCommand {
```

We usually stick to Scala Style Guide, where the recommended indentation is constant
https://docs.scala-lang.org/style/declarations.html#classes

admin/src/main/scala/za/co/absa/spline/arango/AutoClosingArangoManagerProxy.scala
```diff
@@ -16,36 +16,39 @@
...
-      case Failure(e) => Future.failed(e)
-      case Success(v) => v
-    }) andThen {
-      case _ => dbFacade.destroy()
+object AutoClosingArangoManagerProxy {
+
+  def create(
+              managerProvider: ArangoDatabaseAsync => ArangoManager,
+              arangoFacadeProvider: () => ArangoDatabaseFacade)
+            (implicit ex: ExecutionContext): ArangoManager = {
```
cerveada: indentation
https://docs.scala-lang.org/style/indentation.html#methods-with-numerous-arguments

arangodb-foxx-services/src/main/services/prune-database.ts
wajda: I would give it a try as my logic tells me that it would be more correct approach. Logging is not important here IMO.

wajda (dismissed):

admin/src/main/scala/za/co/absa/spline/admin/DateTimeUtils.scala
dk1844: I will be honest, this part I have taken from the original PR https://github.com/AbsaOSS/spline/pull/762
@wajda, do you remember why the parses is used there?

cerveada: Yeah, Alex told me later. He was working on some changes.

admin/src/main/scala/za/co/absa/spline/admin/commands.scala
dk1844: Thanks, changed as suggested.

admin/src/main/scala/za/co/absa/spline/arango/AutoClosingArangoManagerProxy.scala
dk1844: Thanks, changed as suggested.

admin/src/main/scala/za/co/absa/spline/admin/DateTimeUtils.scala
dk1844: So ok as is?

arangodb-foxx-services/src/main/services/prune-database.ts
dk1844: Actually, I tend to think that logging here is crucial. For large pruning, the individual parts can take minutes/tens of minutes and without logging one is blind on what is happening in the DB -- it could run for hours without any sign of the current state.

Here, I would strive to keep it as is for now.

admin/src/main/scala/za/co/absa/spline/admin/DateTimeUtils.scala
```diff
@@ -0,0 +1,60 @@
...
+    .optionalEnd()
+    .parseDefaulting(ChronoField.HOUR_OF_DAY, 0)
+    .toFormatter();
+
+  def parseZonedDateTime(s: String, defaultZoneId: ZoneId = ZoneId.systemDefault): ZonedDateTime =
+    Try {
+      val ZonedDateTimeRegexp(ldt, tzOffset, tzId) = s
+      val maybeTzIds = Seq(tzId, tzOffset).map(Option.apply)
+
+      require(!maybeTzIds.forall(_.isDefined), "Either timezone ID or offset should be specified, not both")
```
cerveada, line 49: `ZonedDateTime.parse` allows both offset and name at the same time . I think it's good idea to accept all inputs that are valid for `ZonedDateTime.parse`

admin/src/main/scala/za/co/absa/spline/admin/DateTimeUtils.scala
cerveada: No. Sorry, my previous message was bit confusing. We discussed the formatter with Alex and he was planning to do some changes in this class.

admin/src/main/scala/za/co/absa/spline/admin/DateTimeUtils.scala
cerveada: The issue with allowing only **geographical zone id** is that it is ambiguous. Not unique. 

```
2022-10-30T02:30:00[Europe/Prague] is both:
2022-10-30T02:30+02:00[Europe/Prague]
2022-10-30T02:30+01:00[Europe/Prague]
```

This is the day and hour of change from summer time to winter time:
```scala
val start = ZonedDateTime.parse("2022-10-30T02:30:00+02:00[Europe/Prague]")
val end = start.plusHours(1)
// start: java.time.ZonedDateTime = 2022-10-30T02:30+02:00[Europe/Prague]
// end: java.time.ZonedDateTime =   2022-10-30T02:30+01:00[Europe/Prague]
```

Another interesting behaviour (not so relevant for us though) is this:
```scala
val start = ZonedDateTime.parse("2022-10-30T02:30:00+02:00")
val end = start.plusHours(1)
// start: java.time.ZonedDateTime = 2022-10-30T02:30+02:00
// end: java.time.ZonedDateTime   = 2022-10-30T03:30+02:00
```
When no geographical zone id is provided the ZonedDateTime will stay in the same offset.

cerveada: We agreed to solve this as part of other ticket #1139

cerveada (approved):

PR #1132, opened 2022-11-09
spline #1131 Re-implement SPLINE::IS_VISIBLE_FROM_TX() function to pu…

### LINKED ISSUE #1131: Re-implement SPLINE::IS_VISIBLE_FROM_TX() function to pure AQL
User-defined functions and other functions that can only run on Coordinator nodes in a cluster are not supported in PRUNE expressions. For the upcoming 3.11, the following change has been made to reject queries with invalid PRUNE conditions:

> **Prevent moving of unsafe vertex or edge filters into the traversal**
Previously any vertex or edge filter condition qualified to be moved into the traversal. This is not safe if the condition is executed on DB servers. For example, on DB servers it is not possible to execute JavaScript user-defined functions, or some other AQL functions that only work on coordinators.

See arangodb/arangodb#16622

_Originally posted by @Simran-B in https://github.com/arangodb/arangodb/issues/17291#issuecomment-1304634272_

wajda:
…re AQL

fixes #1131

1. Replace the `SPLINE::IS_VISIBLE_FROM_TX()` user defined function with inline AQL expressions

The PR is still WIP because of missing unit tests. The included in the PR `aql-gen-helper.spec.ignore.ts` test requires mocks for `@arangodb` module, particularly the `aql` string interpolator that is used for AQL code generation. I need to figure out how to either mock it properly, or use a real embedded database instance to run tests again.


wajda:
@korel-san , any idea how to solve the unit test issue better?

cerveada (approved):

PR #1148, opened 2023-01-04
spline #1139 DateTimeUtils inconsistencies with default ZonedDateTime

### LINKED ISSUE #1139: DateTimeUtils inconsistencies with default ZonedDateTime parsing.
parsing in `admin/src/main/scala/za/co/absa/spline/admin/DateTimeUtils.scala`

`ZonedDateTime.parse` allows both offset and name at the same time . I think it's good idea to accept all inputs that are valid for `ZonedDateTime.parse`

The issue with allowing only **geographical zone id** is that it is ambiguous. Not unique. 

```
2022-10-30T02:30:00[Europe/Prague] is both:
2022-10-30T02:30+02:00[Europe/Prague]
2022-10-30T02:30+01:00[Europe/Prague]
```

This is the day and hour of change from summer time to winter time:
```scala
val start = ZonedDateTime.parse("2022-10-30T02:30:00+02:00[Europe/Prague]")
val end = start.plusHours(1)
// start: java.time.ZonedDateTime = 2022-10-30T02:30+02:00[Europe/Prague]
// end: java.time.ZonedDateTime =   2022-10-30T02:30+01:00[Europe/Prague]
```

Another interesting behaviour (not so relevant for us though) is this:
```scala
val start = ZonedDateTime.parse("2022-10-30T02:30:00+02:00")
val end = start.plusHours(1)
// start: java.time.ZonedDateTime = 2022-10-30T02:30+02:00
// end: java.time.ZonedDateTime   = 2022-10-30T03:30+02:00
```
When no geographical zone id is provided the ZonedDateTime will stay in the same offset.

- `defaultZoneId: ZoneId` parameter is not use at all

wajda:
fixes #1139

Adjusted the `parseZonedDateTime` utility method to allow specifying both timezone Offset and timezone geo ID.
The behavior of this method is now aligned with the logic described in the `ZonedDateTime.ofLocal()`

> The local date-time is resolved to a single instant on the time-line. This is achieved by finding a valid offset from UTC/Greenwich for the local date-time as defined by the rules of the zone ID.
In most cases, there is only one valid offset for a local date-time. In the case of an overlap, where clocks are set back, there are two valid offsets. If the preferred offset is one of the valid offsets then it is used. Otherwise the earlier valid offset is used, typically corresponding to "summer".
In the case of a gap, where clocks jump forward, there is no valid offset. Instead, the local date-time is adjusted to be later by the length of the gap. For a typical one hour daylight savings change, the local date-time will be moved one hour later into the offset typically corresponding to "summer".

### DIFF
--- a/admin/src/main/scala/za/co/absa/spline/admin/DateTimeUtils.scala
+++ b/admin/src/main/scala/za/co/absa/spline/admin/DateTimeUtils.scala
@@ -16,13 +16,16 @@
 
 package za.co.absa.spline.admin
 
+import org.slf4s.Logging
+
 import java.time.format.{DateTimeFormatter, DateTimeFormatterBuilder}
 import java.time.temporal.ChronoField
-import java.time.{ZoneId, ZonedDateTime}
-import scala.util.{Failure, Success, Try}
+import java.time.{LocalDateTime, ZoneId, ZoneOffset, ZonedDateTime}
+import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
+import scala.util.{Failure, Success, Try}
 
-object DateTimeUtils {
+object DateTimeUtils extends Logging {
 
   private val ZonedDateTimeRegexp = (s"" +
     "^" +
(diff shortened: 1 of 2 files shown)


cerveada (approved):
LGTM

PR #1156, opened 2023-01-13
Feature/spline 1111 separate search idx update

### LINKED ISSUE #1111: Decompose lineage ingestion and search index update processes
## Problem
- At the moment _Consumer API_ relies on certain redundant information that is stored in the _Progress_ and _DataSource_ nodes. It makes the _Producer_ services logically dependent on the _Consumer_ ones, which is an undesirable dependency.
- Also, the info stored in the _DataSource_ is transient and is updated every time a new adjacent edge is created. The update operation is not covered with the application level transaction and must be avoided.

## Proposal
Decouple _Consumer_ related parts of the lineage persistence process into a separate service according to the CQRS pattern. Tha is, the second service would listen and react on events produced by the _Producer_ service, like "execution plan inserted", " execution event inserted" etc. The events would be sent at a very last stage of the _Producer_ transaction and delivered in the "at-least-once" manner.

#### Possible implementation
ArangoDB queues

wajda:
fixes #1111 

Mostly refactoring of the Foxx services with the focus on separation of the lineage graph persistent logic and updating certain redundant properties that are only used by the Consumer service

Introduced the transaction state publish-subscribe mechanism


arangodb-foxx-services/package.json
```diff
@@ -10,8 +10,9 @@
     "test": "jest",
     "lint": "eslint src",
     "foxx:deploy": "npm run build && npm run foxx:uninstall && npm run foxx:install",
+    "foxx:deploy-dev": "npm run foxx:deploy --extra_args='--dev'",
     "foxx:uninstall": "cross-env-shell foxx uninstall --server $npm_config_server --database $npm_config_database /spline",
```
korel-san, line 14: @wajda maybe $npm_config_extra_args was missed for foxx:uninstall script?

cerveada (dismissed):

arangodb-foxx-services/src/main/index.ts
```diff
@@ -17,5 +17,7 @@
 import router from './routes'
 import { context } from '@arangodb/locals'
 
+import './modules'
```
korel-san: @wajda I do recommend decreasing Doorway Effect (the best practice is prefer less amount of subfolders)

arangodb-foxx-services/tests/aql/example.spec.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import myCoolFunction = require('../../aql/example.func')
+import myCoolFunction = require('../../src/aql/example.func')
```
korel-san, line 17: @wajda is this test example really needed?

arangodb-foxx-services/package.json
wajda: no, this is correct. Extra args are only applicable to the `foxx:install` script.

arangodb-foxx-services/src/main/utils/logger.ts
```diff
@@ -19,6 +19,10 @@ import { context } from '@arangodb/locals'
 
 const isDebugEnabled: boolean = context.isDevelopment
```
korel-san, line 20: @wajda I would recommend using standard packages for logging https://npmcompare.com/compare/bunyan,loglevel,npmlog,pino,winston

arangodb-foxx-services/tests/aql/example.spec.ts
wajda: well, I would say yes. There is currently no user defined AQL function used in Spline, but there used to be, and could be in the future. And since AQL UDFs are deployed using a separate procedure, we have created a set of scripts and configuration to make this process as simple as merely dropping a file in a source folder. There is however a couple of conventions that must be followed in order for the UDF and the build process to work correctly, so a clear example has to be placed somewhere. And this is the place. We could simply remove this UDF build part along with the UDF source folder due to redundancy, and we'll probably do it later at some point. But for now I'd like to keep it around and ready just in case we decide to create another UDF in the project.

arangodb-foxx-services/src/main/utils/logger.ts
wajda: I wonder if anything of those can be used in Foxx, which I doubt, but I'll take a look. Thanks for the hint.

arangodb-foxx-services/src/main/index.ts
wajda: Less? `modules` is a actually a 1st level folder under the Foxx service source root, which is `/arangodb-foxx-services/src/main` in the project structure.

arangodb-foxx-services/src/main/services/txm/index.ts
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+import { TxManagerImpl } from './tx-manager-impl'
+import { SubscribableTxManagerDecorator } from './subcribable-tx-manager-decorator'
+
+
+export const TxManager = new SubscribableTxManagerDecorator(new TxManagerImpl())
```
korel-san, line 22: @wajda is it really necessary to have 2 TxManager? I didn't find any place with the usage of TxManagerImpl, except here. Do you really expect that we will have another implementation of the TxManager interface? 
Furthermore, subscription is really similar to BehaviourSubject. Maybe it would be more accurate if we had only one BehaviourSubject in the TxManagerImpl, to which anyone could subscribe from just one TxManager instance.

korel-san (dismissed):

arangodb-foxx-services/src/main/index.ts
korel-san: import './modules' 
> `./modules/index.ts`: import './consumer-index-update' 
>> `./modules/consumer-index-update/index.ts`: import './data-source.last-write.update' 
>>> `./modules/consumer-index-update/data-source.last-write.update.ts`

wajda: What's you suggestion?
The idea here is this - we have an core services, and a set of optional loosely coupled modules that suppose to subscribe for system events and do their work not requiring the core part of the system to be aware of their existence. But somehow those modules have to be called for the first time to be able to register themselves. My solution to this is only import `modules/` folder in the core part of the app, and then the rest of the modules would be registered via imports in that umbrella `index.ts`. So when a new module is added to or deleted from the system, all that needs to be done is just to create/delete a module folder and add/remove the respective `import` statement from the `modules/index.ts`. This way the core system is kind of remains unaware of the modules, while technically still execute them indirectly. How would you do it without this indirection layer?

wajda: webpack?

arangodb-foxx-services/src/main/utils/logger.ts
wajda: Glanced through all the libraries and... No.
They all are designed for a pure Nodejs or Browser logging, supporting additional levels, fancy formatting, alternative streams, log rotations etc etc etc. But nothing of those feature are required or even possible in this app. All logging in Foxx is done via `Console`, that is then routed to the internal ArangoDB logging system which then does all logging management centrally. Unfortunately there is no Foxx specific appender for any of those frameworks (or at all), so we'd have to write our own appender anyway. All of that above tells me that using any external logging library for this app is simply unnecessary over engineering.

arangodb-foxx-services/src/main/services/txm/index.ts
wajda: I'm not sure if `rxjs` is compatible with the limited Foxx environment (I see `await` is used in some places of the source code, but Foxx is strictly synchronous and async logic would likely either not work or behave unpredictably). But even if it works, how is `Subject` (not `BehaviorSubject` btw) better than `EventEmmiter` in this application? To me there is no diff in the current use case, but `events` library is provided by Foxx while `rxjs` is an external one.

wajda: Regarding the `TxManager` interface, it has two implementations: `TxManagerImpl` and `SubscribableTxManagerDecorator`. The reason for not combining them in one is _separation of concerts_. Each class implements it's own logic and then both are cohesively connected via the _Decorator_ patern.

arangodb-foxx-services/src/main/index.ts
wajda: I tried to redo it via Webpack entry configuration and I must say I like it much more. Now modules are purely auto discoverable and completely decoupled (dependency-wise) from the core. And the project structure is less nested and nicer.

```javascript
entry: {
    'index': [
        ...Glob.sync(`./src/main/index.ts`),
        ...Glob.sync(`./src/modules/**/index.ts`),
    ],
    'scripts/setup': './src/scripts/setup.ts',
    'scripts/teardown': './src/scripts/teardown.ts',
    ...
}
```

korel-san (approved):

PR #1187, opened 2023-03-02
spline #1170 starting the query from execution plan

### LINKED ISSUE #1170: `/execution-plans` works incorrectly
Related to the feature #1142 that was not implemented correctly.

The `GET /execution-plans` must return *unique* execution plans, and the AQL query should start with scanning `executionPlan` collection respectively. The following comment from the `ExecutionPlanRepositoryImpl.scala:122` is misleading and is likely incorrect:

```scala
    // cannot use:
    //        FOR execPlan IN executionPlan
    //          FOR prog IN 1 OUTBOUND execPlan progressOf
    // because CURRENT's context would be `progress`, not `executionPlan`
```

Plus, the `/execution-plans` REST endpoint should accept more GET parameters to allow for search & filter similarly to `/execution-events`. At a minimum it should accept `searchTerm`, `append` and the `label` parameters.

cerveada:
- not returning graph

fixes #1170


wajda (approved):
LGTM

PR #1194, opened 2023-03-22
spline #1193 Admin: add `--dry-run` option

### LINKED ISSUE #1193: Admin: add `--dry-run` option

wajda:
fixes #1193

### DIFF
--- a/admin/src/test/scala/za/co/absa/spline/admin/AdminCLISpec.scala
+++ b/admin/src/test/scala/za/co/absa/spline/admin/AdminCLISpec.scala
@@ -77,9 +77,10 @@ class AdminCLISpec
     val dbOptionCaptor: ArgumentCaptor[DatabaseCreateOptions] = ArgumentCaptor.forClass(classOf[DatabaseCreateOptions])
     val actionFlgCaptor: ArgumentCaptor[OnDBExistsAction] = ArgumentCaptor.forClass(classOf[OnDBExistsAction])
     val sslCtxCaptor: ArgumentCaptor[Option[SSLContext]] = ArgumentCaptor.forClass(classOf[Option[SSLContext]])
+    val dryRunModeCaptor: ArgumentCaptor[Boolean] = ArgumentCaptor.forClass(classOf[Boolean])
 
     (when(
-      arangoManagerFactoryMock.create(connUrlCaptor.capture, sslCtxCaptor.capture))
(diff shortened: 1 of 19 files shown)


persistence/src/main/scala/za/co/absa/spline/persistence/DryRunnable.scala
```diff
@@ -0,0 +1,36 @@
...
+    else code
+  }
+
+  def unlessDryRun[T <: AnyRef](code: => T): T = {
+    unlessDryRun(code, null).asInstanceOf[T]
+  }
+
+  def unlessDryRunAsync[T <: Future[U], U](code: => U): U = {
+    unlessDryRun(code, Future.successful(null)).asInstanceOf[U]
+  }
```
cerveada: What is `T` needed for? Seems to be unused. Should some of the `U`s be `T`s?

admin/src/test/scala/za/co/absa/spline/admin/AdminCLISpec.scala
```diff
@@ -123,6 +124,14 @@ class AdminCLISpec
...
     }
 
+    it should "when called with option --dry-run, set dryRun mode to true" in {
+      val output = captureStdOut {
+        cli.exec(Array("db-exec", "arangodbs://foo/bar", "--dry-run"))
+      }
+      dryRunModeCaptor.getValue should be(true)
+      output should include("Dry-run mode activated")
+    }
+
```
cerveada, line 134: Some "integration" test validating that when dry run is set there is no interaction with the parts of the code that should not be called, could be also useful. Maybe in different filet though.

persistence/src/main/scala/za/co/absa/spline/persistence/DryRunnable.scala
wajda: Ah, that's a typo, thanks :)

admin/src/test/scala/za/co/absa/spline/admin/AdminCLISpec.scala
wajda: I thought about it, but it has to be an integration test with a real ArangoDB. Unfortunately capturing this behavior with mocks is impossible. Due to the nature of the ArangoDB driver there is no one place where you can intercept all write operations, or isolate all write method calls. The guard conditions are spread almost freely over the Admin codebase.
I'll see what I can do.

wajda: added ArangoDb test container, fixtures and a couple of integration tests

cerveada (approved):

PR #1195, opened 2023-03-22
spline #1191 Admin: print a backup reminder before database migration

### LINKED ISSUE #1191: Admin: print a backup reminder before database migration
Before starting a migration procedure, print a bold message suggesting creating a database backup.
Something like the following:
```shell
admin-cli db-upgrade ....

**************************************************************
**                      WARNING !                           **
** This operation is not reversible! Please make sure       **
** you created a full database backup before continue       **
**                                                          **
** Instruction how to make backups can be found here:       **
** https://www.arangodb.com/docs/stable/backup-restore.html **
**************************************************************

I confirm that I created a database backup: [y/N]
```

The message and the prompt should only appear when CLI is executed in the interactive mode.

wajda:
Fixes #1191 

Result:

image


cerveada:
What if the upgrade is run via script, not manually, will this cause issue in such a case? Should there be some `-y` option to agree automatically?

wajda:
> What if the upgrade is run via script, not manually, will this cause issue in such a case? Should there be some `-y` option to agree automatically?

All the user interactions are enabled when there is a console. In scripts, or e.g. when you run it from IDEA etc when the console is not attached to the process the above message won't appear.
The main intention for this feature is just to remind the user about the necessity of database backup. In script it would be too late anyway.

wajda:
@korbelm , do you think we should print it even in the non-interactive mode, effectively always requiring `-y` option in addition to `db-upgrade` while running it from scripts?

korbelm:
i would keep `-y` even for automated run (other sw do) as it should make users more aware of potential risks, from my point of view there is no need to have separate behaviour

wajda:
Added the `--non-interactive` flag suggested above. I didn't use `-y` as it implies "yes", but there are also other prompts in the Admin CLI, not just yes/no.

The updated logic is the following:

- If the `--non-interactive` flag is used, regardless of the console availability the process just carries on after printing the corresponding warning message.

image

- If the `--non-interactive` is NOT used then the behavior depends on the console availability. If it's available, the command acts interactively

image

- otherwise it prints an error message instructing the user to switch into the non-interactive mode.

image

cerveada (approved):

